PR #15, opened 2016-10-23
added colors from style guide

doctor3w:
Added the style guide colors for app-wide use. 

The pink color may need to be updated because it wasn't correct on the style guide but I color metered it to get the RGBs and then hex from it. So it might not be exact.


anniexcheng:
Have Eileen check over the colors and if she says they're fine, then you can merge it in :)

Issue #10, opened 2016-10-03
Better download of large podcast files


mrkbryn:
Since podcast files are generally large and we can reasonably start playing audio after downloading a minute or more of the podcast (allowing us to continue downloading the rest of the podcast in the background), we should find a way to do this with NSURLRequest or something.

This may not be able to be implemented until we have JSON responses for podcasts telling us how large the podcast audio file is.

Player should pause if unable to download the rest of podcast.

mrkbryn:
Closed with #16

PR #16, opened 2016-11-06
Improved Player Class

mrkbryn:
Addresses #10 and iterates on the Player. Player class now uses AVPlayer instead of AVAudioPlayer so that we can download large files and begin playback before the entire file has been downloaded. While the audio file is being prepared, togglePlay controls whether the Player singleton will autoplay once the file has buffered enough.

I've designed the Player so that playback can be controlled from anywhere inside the app. This will help later when we implement interactions where a user clicks on a podcast cell intending to play the track. In order to grab control of the playback of the Player, use the `Player.sharedInstance` singleton to call `prepareToPlay(url: url)` and then call `togglePlaying()` to set the player to auto play once the track has buffered. In order to observe the state changes and playback progress of the player to make UI changes, set `Player.sharedInstance.delegate` to `self`.

Also included in this PR is a dynamic player UI that adjusts time labels, images, and slider position based on the progress of the Player instance.


Podcast/Extensions.swift
```diff
@@ -43,3 +44,18 @@ extension UIFont {
     @nonobjc static let discoverTableViewCellDefaultFontAttributes = [NSForegroundColorAttributeName: UIColor.podcastGrayLight, NSFontAttributeName: UIFont(name: ".SFUIText-Medium", size: 10.0)!]
 
 }
+
+extension CMTime {
+    var durationText:String {
+        let totalSeconds = Int(self.seconds)
+        let hours = totalSeconds / 3600
```
dennisfedorko, line 53: need to use floats here, as you're currently dividing two Ints which will give you a rounded Int instead of a double

Podcast/Player.swift
```diff
@@ -9,59 +9,189 @@
...
     
-    // MARK: - AVAudioPlayerDelegate
+    /// Skips a variable amount of seconds in the current audio track. Skips forward or backward based on a positive or negative seconds value.
+    /// - Parameters:
+    ///   - seconds: the amount of time in seconds to skip. If negative, skips backward.
+    func skip(seconds: Float64) {
+        if let player = currentAVPlayer {
+            let newTime = CMTimeAdd(player.currentTime(), CMTimeMakeWithSeconds(seconds, Int32(NSEC_PER_SEC)))
+            player.seek(to: newTime)
+            self.delegate?.playerDidUpdateTime()
```
dennisfedorko: don't need self here

Podcast/Player.swift
```diff
@@ -9,59 +9,189 @@
...
+        shouldAutoPlay = false
         super.init()
     }
     
-    var isPlaying: Bool {
-        return player?.isPlaying ?? false
+    var playerStatus: PlayerStatus {
+        didSet {
+            if oldValue != playerStatus {
+                self.delegate?.playerDidChangeState()
```
anniexcheng: don't need self here

Podcast/Player.swift
```diff
@@ -9,59 +9,189 @@
...
+    private var playerContext: UnsafeMutableRawPointer?
+    private var currentAVPlayer: AVPlayer? {
+        didSet {
+            oldValue?.pause()
+        }
+    }
+    private var currentURL: URL? {
+        didSet {
+            if oldValue != currentURL {
+                prepareToPlay(url: currentURL!)
```
anniexcheng: Is the currentURL guaranteed to not be nil? It's usually better to use ? since you defined currentURL as an optional value

Podcast/Player.swift
```diff
@@ -9,59 +9,189 @@
...
+        switch playerStatus {
+        case .empty:
+            break
+        case .preparingToPlay:
+            shouldAutoPlay = !shouldAutoPlay
+        case .playing:
+            pause()
+        case .paused:
+            play()
+        case .failed:
```
anniexcheng: let's combine this line with lines 98-99:
`case .empty, .failed:
     break`

doctor3w:
Looks good to me, just make those changes above.

Podcast/Player.swift
```diff
@@ -9,59 +9,189 @@
...
+        }
+        
+        if progress < 0.0 {
+            setProgress(progress: 0.0)
+        } else if progress > 1.0 {
+            setProgress(progress: 1.0)
+        }
+        
+        let duration = currentAVPlayer?.currentItem?.duration
+        let newTime = progress * Float((duration?.seconds)!)
```
anniexcheng: duration may be nil if currentAVPlayer is nil so to safeguard against that, do 
`if let duration = currentAVPlayer?.currentItem?.duration ...`

anniexcheng:
See @lucasderraugh's comments below

Podcast/Player.swift
```diff
@@ -9,59 +9,189 @@
...
-                    guard let s = self else { return }
-                    guard let data = data else { return }
-                    
-                    s.player = try? AVAudioPlayer(data: data)
-                    s.player?.prepareToPlay()
-                    s.currentTask = nil
-                    s.player?.play()
-                }
-                currentTask?.resume()
+    var delegate: PlayerDelegate?
```
lucasderraugh: Delegate should be weak

Podcast/Player.swift
```diff
@@ -9,59 +9,189 @@
...
+    case paused
+    case failed
+}
+
+/// Protocol for the singleton Player to observe changes in the Player's state and time changes
+protocol PlayerDelegate {
+    /// Called when the Player's state changes. Override to observe when the state is changed.
+    func playerDidChangeState() -> Void
+    /// Called when the Player has made playback progress. Override to observe when the Player makes progress on playback.
+    func playerDidUpdateTime() -> Void
```
lucasderraugh: Unnecessary `-> Void`

Podcast/Extensions.swift
```diff
@@ -43,3 +44,18 @@ extension UIFont {
     @nonobjc static let discoverTableViewCellDefaultFontAttributes = [NSForegroundColorAttributeName: UIColor.podcastGrayLight, NSFontAttributeName: UIFont(name: ".SFUIText-Medium", size: 10.0)!]
 
 }
+
+extension CMTime {
+    var durationText:String {
```
lucasderraugh: Keep those style guidelines, space after colon.

Podcast/Player.swift
```diff
@@ -9,59 +9,189 @@
...
     
+    /// Prepares the player to play a new track from a URL. If the Player is in the .preparingToPlay state, this has no effect on current playback.
+    /// - Parameters:
+    ///   - url: The url to fetch and play from.
+    func prepareToPlay(url: URL) {
+        if currentURL != url || playerStatus != .preparingToPlay {
+            // create a new player if this is a new URL or we aren't currently preparing a URL
+            playerStatus = .preparingToPlay
+            currentAVPlayer = AVPlayer(url: url)
+            currentAVPlayer?.addObserver(self, forKeyPath: #keyPath(AVPlayer.status), options: [.old, .new], context: &playerContext)
```
lucasderraugh: KVO warnings going off in my head! Where are you removing this observer? (The answer should probably be on deinit as well as anytime you change up this `currentAVPlayer`)

Podcast/Player.swift
lucasderraugh: Or `default` case them :)

Podcast/Player.swift
```diff
@@ -9,59 +9,189 @@
...
+        }
+    }
     
     
-    func audioPlayerDidFinishPlaying(_ player: AVAudioPlayer, successfully flag: Bool) {
-        pause()
+    /// Returns the progress for the Player's current track.
+    /// - Returns: a Float in the range [0.0,1.0] specifying what the current location of playback is relative to the entire track.
+    func getProgress() -> Float {
+        if playerStatus == .empty || playerStatus == .failed || currentAVPlayer == nil || currentAVPlayer?.currentItem == nil {
```
lucasderraugh: Redundant case `currentAVPlayer == nil` since the optional chaining on the next condition will yield nil if the player is nil.

Podcast/Player.swift
```diff
@@ -9,59 +9,189 @@
...
+    ///   - progress: the value to set where the Player should begin playback. Should be in range [0.0,1.0], but will be set to 0.0 if less than 0.0 and 1.0 if greater than 1.0
+    func setProgress(progress: Float) {
+        if playerStatus == .empty || playerStatus == .failed {
+            return
+        }
+        
+        if progress < 0.0 {
+            setProgress(progress: 0.0)
+        } else if progress > 1.0 {
+            setProgress(progress: 1.0)
```
lucasderraugh: Why would we call the function again? Just clamp the value (min+max) around 0-1 and use that for the rest of the computation.

Podcast/Player.swift
```diff
@@ -9,59 +9,189 @@
...
+        let newTime = progress * Float((duration?.seconds)!)
+        currentAVPlayer?.seek(to: CMTimeMakeWithSeconds(Float64(newTime), Int32(NSEC_PER_SEC)))
+    }
+    
+    /// - Returns: the time passed if there is a current item in the Player. nil if no item
+    func getTimePassed() -> CMTime? {
+        if let currentItem = currentAVPlayer?.currentItem {
+            return currentItem.currentTime()
+        }
+        return nil
```
lucasderraugh: One liner this sucker into `return currentAVPlayer?.currentItem?.currentTime()`

Same with `getTimeLeft()` below

Podcast/Extensions.swift
mrkbryn: I think we don't really need the float/double precision for calculating the display time. The durationText is just going to be used for displaying on the Player UI in the format `h:mm:ss` or `mm:ss` so we never need to calculate this time in terms of fractions of a second.

Maybe that wasn't clear in the CMTime extension, so I just added a line of documentation of it to clarify.

Podcast/Player.swift
mrkbryn: oops 😆 I now add an observer in `play()` and remove it in `pause()` since we don't need to observe the time when it's paused. also added a `deinit` to remove it there

PR #33, opened 2017-02-18
Create custom tab bar

dennisfedorko:
Please review to make sure everything makes sense

Obviously we can change things around, waiting on that floating player view to be implemented to add that in, but this is what I came up with, its by no means perfect, but allows us to do some custom behavior/animations that we couldn't do with the vanilla UITabBarController.


Podcast/TabBarController.swift
```diff
@@ -0,0 +1,149 @@
...
+            
+            let blurEffect = UIBlurEffect(style: UIBlurEffectStyle.extraLight)
+            let blurEffectView = UIVisualEffectView(effect: blurEffect)
+
+            blurEffectView.frame = tabBarContainerView.bounds
+            blurEffectView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+            
+            tabBarContainerView.addSubview(blurEffectView)
+            
+        } else {
```
doctor3w: Do you need this else if you set it above before the if-statement?

Podcast/TabBarController.swift
```diff
@@ -0,0 +1,149 @@
...
+            tabBarButtons.append(newTabBarButton)
+            
+            xOffset += tabBarButtonWidth
+        }
+        
+    }
+    
+    func didPressTabBarButton(tabBarButton: UIButton) {
+        
+        guard let tabBarButtonIndex = tabBarButtons.index(of: tabBarButton) else { return }
```
doctor3w, line 94: Why not use button tag as index rather than search through an array?

doctor3w (approved):

Podcast/TabBarController.swift
dennisfedorko: agreed

Podcast/TabBarController.swift
dennisfedorko: im not a fan of using tags, its a 4 button array, so search is ok

doctor3w: Alright sounds good

PR #50, opened 2017-03-02
making card and episode cell tag buttons

natashaarmbrust:


Podcast/CardTableViewCell.swift
```diff
@@ -305,7 +303,51 @@ class CardTableViewCell: UITableViewCell {
         
     }
     
-
+    func setupTagButtons(card: EpisodeCard) {
```
dennisfedorko: why do you need to pass the entire card in? why not an array of tags?

Podcast/Episode.swift
```diff
@@ -22,7 +22,7 @@ class Episode: NSObject {
     var largeArtworkImage: UIImage!
     var mp3URL : URL?
     var time: Double = 0
-    var tags : [String] = []
+    var tags : [Tag] = []
```
dennisfedorko: tags: [Tag], no space before :

Podcast/Episode.swift
```diff
@@ -22,7 +22,7 @@ class Episode: NSObject {
     var largeArtworkImage: UIImage!
     var mp3URL : URL?
```
dennisfedorko: mp3URL: URL?

Podcast/Episode.swift
```diff
@@ -22,7 +22,7 @@ class Episode: NSObject {
...
     var time: Double = 0
-    var tags : [String] = []
+    var tags : [Tag] = []
     var nRecommended : Int = 0
```
dennisfedorko: rename plz

Podcast/EpisodeTableViewCell.swift
```diff
@@ -304,6 +282,49 @@ class EpisodeTableViewCell: UITableViewCell {
         cardID = episode.id
     }
     
+    func setupTagButtons(episode: Episode) {
```
dennisfedorko: why is this in cardtableviewcell and episrodetableviewcell?

Podcast/Episode.swift
natashaarmbrust: it will be in my next PR when i work on the models but for now I am leaving it

Podcast/EpisodeTableViewCell.swift
natashaarmbrust: because they are two different cells ... with different views, etc. Where would you suggest I put it?

Podcast/CardTableViewCell.swift
natashaarmbrust: i could just do an array of tags yup!

PR #58, opened 2017-03-07
Natasha/subscriptions

natashaarmbrust:


Podcast/SubscriptionsCollectionViewFlowLayout.swift
```diff
@@ -0,0 +1,26 @@
...
+//  SubscriptionsCollectionViewFlowLayout.swift
+//  Podcast
+//
+//  Created by Natasha Armbrust on 3/6/17.
+//  Copyright © 2017 Cornell App Development. All rights reserved.
+//
+
+import UIKit
+
+class SubscriptionsCollectionViewFlowLayout: UICollectionViewFlowLayout {
```
dennisfedorko: you don't need this class, you can make the subscriptionsviewcontroller class implement the collectionviewflowlayoutdelegate protocol and implement all these as delegate functions.

Podcast/SubscriptionsViewController.swift
```diff
@@ -0,0 +1,81 @@
...
+    }
+    
+    //MARK:
+    //MARK: - fetch data
+    //MARK:
+    
+    func fetchSubscriptions() -> [Series] {
+        //dummy data
+        var series: [Series] = []
+        for i in 0..<9{
```
dennisfedorko: 0 ..< 9 {

Podcast/Util.swift
```diff
@@ -0,0 +1,67 @@
...
+import UIKit 
+
+class Util {
+    
+    //format date interval between fromDate to toDate by greatest component between
+    // i.e. if fromDate: March 7, 1989  toDate: March 8, 1989 
+    //      formatDateDifferenceByLargestComponent(fromDate, toDate) -> "1 day ago"
+    //      if fromDate: March 9, 2015  toDate: March 8, 2017
+    //      formatDateDifferenceByLargestComponent(fromDate, toDate) -> "2 years ago"
+    static func formatDateDifferenceByLargestComponent(fromDate: Date, toDate: Date) -> String {
```
dennisfedorko: looks to me like this should be an extension on Date rather than making this Util class

Podcast/Util.swift
```diff
@@ -0,0 +1,67 @@
...
+        
+        let calendar = Calendar.current
+        
+        guard let startDays = calendar.ordinality(of: .day, in: .era, for: fromDate) else { return "" }
```
dennisfedorko: combine all these guards

Podcast/Util.swift
```diff
@@ -0,0 +1,67 @@
...
+        let yearsBetween = endYear - startYear
+        
+        if yearsBetween > 0 {
+            if yearsBetween < 2 {
+                return String(yearsBetween) + " year ago"
+            }
+            return String(yearsBetween) + " years ago"
+        }
+        
+        if monthsBetween > 0 {
```
dennisfedorko: why are there two if monthsBetween > 0 if statements? combine them

Podcast/Util.swift
```diff
@@ -0,0 +1,67 @@
...
+        
+        guard let startMonth = calendar.ordinality(of: .month, in: .era, for: fromDate) else { return "" }
+        guard let endMonth = calendar.ordinality(of: .month, in: .era, for: toDate) else { return "" }
+        let monthsBetween = endMonth - startMonth
+        
+        guard let startYear = calendar.ordinality(of: .year, in: .era, for: fromDate) else { return "" }
+        guard let endYear = calendar.ordinality(of: .year, in: .era, for: toDate) else { return "" }
+        let yearsBetween = endYear - startYear
+        
+        if yearsBetween > 0 {
```
dennisfedorko: if yearsBetween > 0 && yearsBetween < 2 {
   return ....
else {
   return ...
}

Podcast/Util.swift
dennisfedorko: combine if statements to avoid nesting

Podcast/Util.swift
```diff
@@ -0,0 +1,67 @@
...
+        
+        if monthsBetween > 0 {
+            if monthsBetween < 2 {
+                return String(monthsBetween) + " month ago"
+            }
+            return String(monthsBetween) + " months ago"
+        }
+        
+        if weeksBetween > 0 {
+            if weeksBetween < 2 {
```
dennisfedorko: combine if statements to avoid nesting

Podcast/Util.swift
```diff
@@ -0,0 +1,67 @@
...
+                return String(weeksBetween) + " week ago"
+            }
+            return String(weeksBetween) + " weeks ago"
+        }
+        
+        if daysBetween > 0 {
+            if daysBetween < 2 {
```
dennisfedorko: combine if statements to avoid nesting

Podcast/SubscriptionsViewController.swift
```diff
@@ -0,0 +1,81 @@
...
+        subscriptionsCollectionView.backgroundColor = .podcastWhiteDark
+        subscriptionsCollectionView.delegate = self
+        subscriptionsCollectionView.dataSource = self
+        subscriptionsCollectionView.showsVerticalScrollIndicator = false
+        view.addSubview(subscriptionsCollectionView)
+        
+        subscriptions = fetchSubscriptions()
+    }
+
+    override func didReceiveMemoryWarning() {
```
dennisfedorko: delete this

Podcast/SubscriptionsViewController.swift
```diff
@@ -0,0 +1,81 @@
...
+    func numberOfSections(in collectionView: UICollectionView) -> Int {
+        return 1
+    }
+    
+    func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
+        return subscriptions.count
+    }
+    
+    func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
+        let cell = collectionView.dequeueReusableCell(withReuseIdentifier: "SubscriptionsCollectionViewCellIdentifier", for: indexPath) as! SubscriptionCollectionViewCell
```
dennisfedorko: guard let

Podcast/SubscriptionCollectionViewCell.swift
```diff
@@ -0,0 +1,17 @@
...
+//  SubscriptionCollectionViewCell.swift
+//  Podcast
+//
+//  Created by Natasha Armbrust on 3/6/17.
+//  Copyright © 2017 Cornell App Development. All rights reserved.
+//
+
+import UIKit
+
+class SubscriptionCollectionViewCell: RecommendedSeriesCollectionViewCell {
```
dennisfedorko: why is this subclassing recommendedseriescollectionviewcell? isn't this supposed to be completely unrelated to any other cell?

Podcast/SubscriptionsCollectionViewFlowLayout.swift
natashaarmbrust: I personally think this adds clarity and it is what we have done in every other collection view cell but I can change it if necessary

Podcast/Util.swift
natashaarmbrust: Fair i can do that its just pretty specific do I thought a Util would be better

Podcast/Util.swift
natashaarmbrust: but then everything would be nested under the else so the end check (i.e. days) would be super nested

Podcast/SubscriptionCollectionViewCell.swift
natashaarmbrust: Because its literally the exact same cell except for the label text.. I had it as a completely different cell at first then I thought that was super redundant .. If they end up changing we can change it but I think for now it just adds to the codebase for no purpose

Podcast/SubscriptionsViewController.swift
natashaarmbrust: addressed

Podcast/SubscriptionsViewController.swift
natashaarmbrust: addressed

Podcast/SubscriptionsCollectionViewFlowLayout.swift
dennisfedorko: I believe Kevin implemented his cells using the delegate protocol, it makes it a bit more clear because the view controller is responsible for adjusting the collection view. Unless we plan on using this same flow layout elsewhere, I say make it part of the VC

Podcast/Util.swift
dennisfedorko: Util is a very vague class, I'm worried we might just use it as a broom closet for random functions, better to put them as extensions, since this basically takes two dates and outputs a string

Podcast/Util.swift
dennisfedorko: oh I see, nvm

Podcast/Util.swift
natashaarmbrust: addressed put in Date extension

Podcast/SubscriptionCollectionViewCell.swift
dennisfedorko: why not just standardize it by only having one SeriesGridCollectionViewCell class that we can use everywhere we have those series cell blocks.

dennisfedorko: instead of having to subclass, you can just initialize it with an enum: .subscription or .recommended and based on that set the label text

Podcast/SubscriptionsCollectionViewFlowLayout.swift
natashaarmbrust: Addressed

Podcast/Util.swift
natashaarmbrust: addressed

Podcast/SubscriptionCollectionViewCell.swift
natashaarmbrust: sounds good ill make that change now

PR #65, opened 2017-03-08
connect player to feed and bookmark tableview cells

mrkbryn:


Podcast/BookmarkViewController.swift
```diff
@@ -15,6 +15,7 @@ class BookmarkViewController: UIViewController, UITableViewDelegate, UITableView
     ///
     var bookmarkTableView: UITableView!
     var cards: [Card] = []
+    var currentlyPlayingIndexPath: IndexPath?
```
dennisfedorko, line 18: can we change it to currentlyPlayingCardIndex: Int instead?

Podcast/BookmarkViewController.swift
```diff
@@ -44,8 +45,16 @@ class BookmarkViewController: UIViewController, UITableViewDelegate, UITableView
     
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
-        bookmarkTableView.reloadData()
         
+        // check before reloading data whether the Player has stopped playing the currentlyPlayingIndexPath
+        if let indexPath = currentlyPlayingIndexPath {
+            if let card = cards[indexPath.row] as? EpisodeCard {
+                if Player.sharedInstance.currentEpisode?.id != card.episode.id {
```
dennisfedorko: this can all be flattened into one if statement

Podcast/BookmarkViewController.swift
```diff
@@ -44,8 +45,16 @@ class BookmarkViewController: UIViewController, UITableViewDelegate, UITableView
     
     override func viewDidAppear(_ animated: Bool) {
```
dennisfedorko: can this go in viewWillAppear instead?

Podcast/BookmarkViewController.swift
```diff
@@ -101,8 +113,18 @@ class BookmarkViewController: UIViewController, UITableViewDelegate, UITableView
     func cardTableViewCellDidPressPlayPauseButton(cardTableViewCell: CardTableViewCell) {
         guard let cardIndexPath = bookmarkTableView.indexPath(for: cardTableViewCell), let card = cards[cardIndexPath.row] as? EpisodeCard else { return }
         
-        //card.isPlaying = !card.isPlaying
-        //cardTableViewCell.setPlayButtonToState(isPlaying: card.isPlaying)
+        if cardIndexPath == currentlyPlayingIndexPath { return }
+        
+        if let indexPath = currentlyPlayingIndexPath {
+            if let cell = bookmarkTableView.cellForRow(at: indexPath) as? CardTableViewCell {
```
dennisfedorko: flatten into one if statement

Podcast/BookmarkViewController.swift
```diff
@@ -101,8 +113,18 @@ class BookmarkViewController: UIViewController, UITableViewDelegate, UITableView
...
+                cell.setPlayButtonToState(isPlaying: false)
+            }
+        }
+        currentlyPlayingIndexPath = cardIndexPath
+        cardTableViewCell.setPlayButtonToState(isPlaying: true)
+        let appDelegate = UIApplication.shared.delegate as! AppDelegate
```
dennisfedorko: guard let

dennisfedorko: and you can combine it with the guard let above

Podcast/BookmarkViewController.swift
```diff
@@ -101,8 +113,18 @@ class BookmarkViewController: UIViewController, UITableViewDelegate, UITableView
...
+        appDelegate.showPlayer(animated: true)
+        Player.sharedInstance.playEpisode(episode: card.episode)
```
dennisfedorko: what happens if we pause?

Podcast/FeedViewController.swift
```diff
@@ -23,6 +23,7 @@ class FeedViewController: UIViewController, UITableViewDelegate, UITableViewData
     ///
     var feedTableView: UITableView!
     var cards: [Card] = []
+    var currentlyPlayingIndexPath: IndexPath?
```
dennisfedorko, line 26: can we make this currentlyPlayingCard: Int

Podcast/FeedViewController.swift
```diff
@@ -52,8 +53,16 @@ class FeedViewController: UIViewController, UITableViewDelegate, UITableViewData
     
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
-        feedTableView.reloadData()
         
+        // check before reloading data whether the Player has stopped playing the currentlyPlayingIndexPath
+        if let indexPath = currentlyPlayingIndexPath {
+            if let card = cards[indexPath.row] as? EpisodeCard {
```
dennisfedorko: same issues as will feedvc

Podcast/FeedViewController.swift
```diff
@@ -109,8 +121,18 @@ class FeedViewController: UIViewController, UITableViewDelegate, UITableViewData
     func cardTableViewCellDidPressPlayPauseButton(cardTableViewCell: CardTableViewCell) {
         guard let cardIndexPath = feedTableView.indexPath(for: cardTableViewCell), let card = cards[cardIndexPath.row] as? EpisodeCard else { return }
         
-        //card.isPlaying = !card.isPlaying
-        //cardTableViewCell.setPlayButtonToState(isPlaying: card.isPlaying)
+        if cardIndexPath == currentlyPlayingIndexPath { return }
+        
+        if let indexPath = currentlyPlayingIndexPath {
+            if let cell = feedTableView.cellForRow(at: indexPath) as? CardTableViewCell {
+                cell.setPlayButtonToState(isPlaying: false)
```
dennisfedorko: same issues as in feedVC

Podcast/BookmarkViewController.swift
mrkbryn: I think this might complicate things, since I need to call `setPlayButtonToState(isPlaying: false)` on the previously playing cell, which i get through calling `bookmarkTableView.cellForRow(at: indexPath)`. If we change it to an Int, I don't believe there's a function to get a cell from just an Int

mrkbryn: Unless we reload the data for all cells which would be costly if there are many cards in the table view

dennisfedorko: okay thats fine I guess

Podcast/BookmarkViewController.swift
mrkbryn: Do we want to be able to pause by clicking on the cell? I thought the "Now Playing" label implied that this wasn't a typical play/pause button. This is similar to how Spotify functions with cells controlling the starting of playback and the mini player being used to control play/pause/etc

Podcast/BookmarkViewController.swift
mrkbryn: How should I do this? I think I have to unwrap the indexPath before I use it to get the correct card from cards

mrkbryn: nvm do you mean chain multiple `if let` into one line?

Podcast/BookmarkViewController.swift
mrkbryn: Yup, I moved it in the latest commit

PR #72, opened 2017-03-20
Drew/bookmarks cell

doctor3w:
Created the new smaller bookmarks cell to go in the bookmarks view controller.


Podcast/BookmarksTableViewCell.swift
```diff
@@ -0,0 +1,228 @@
...
+import UIKit
+
+protocol BookmarksTableViewCellDelegate: class {
+    
+    func bookmarksTableViewCellDidPressPlayPauseButton(bookmarksTableViewCell: BookmarksTableViewCell)
+    func bookmarksTableViewCellDidPressRecommendButton(bookmarksTableViewCell: BookmarksTableViewCell)
+    func bookmarksTableViewCellDidPressMoreActionsButton(bookmarksTableViewCell: BookmarksTableViewCell)
+}
+
+class BookmarksTableViewCell: UITableViewCell {
```
dennisfedorko: rename to BookmarkTableViewCell

Podcast/BookmarksTableViewCell.swift
```diff
@@ -0,0 +1,228 @@
...
+//  BookmarksTableViewCell.swift
+//  Podcast
+//
+//  Created by Drew Dunne on 3/20/17.
+//  Copyright © 2017 Cornell App Development. All rights reserved.
+//
+
+import UIKit
+
+protocol BookmarksTableViewCellDelegate: class {
```
dennisfedorko: rename to BookmarkTableViewCell

Podcast/BookmarksTableViewCell.swift
```diff
@@ -0,0 +1,228 @@
...
+    
+    func bookmarksTableViewCellDidPressPlayPauseButton(bookmarksTableViewCell: BookmarksTableViewCell)
```
dennisfedorko: rename to bookmark instead of bookmarks everywhere

Podcast/BookmarksTableViewCell.swift
```diff
@@ -0,0 +1,228 @@
...
+        recommendedButton.titleLabel?.font = .systemFont(ofSize: 12)
+        recommendedButton.addTarget(self, action: #selector(didPressRecommendedButton), for: .touchUpInside)
+        addSubview(recommendedButton)
+        
+        moreButton = UIButton(frame: CGRect.zero)
+        moreButton.setImage(#imageLiteral(resourceName: "more_icon"), for: .normal)
+        moreButton.addTarget(self, action: #selector(didPressMoreButton), for: .touchUpInside)
+        addSubview(moreButton)
+        
+        heightConstraint = NSLayoutConstraint(item: contentView, attribute: .height, relatedBy: .equal, toItem: nil, attribute: NSLayoutAttribute.notAnAttribute, multiplier: 1, constant: height)
```
dennisfedorko: don't use NSLayoutConstraints, use frame based layout

Podcast/BookmarksTableViewCell.swift
```diff
@@ -0,0 +1,228 @@
...
+        moreButton.frame = CGRect(x: frame.width - moreButtonWidth - moreButtonRightX, y: frame.height - moreButtonBottomY - moreButtonHeight, width: moreButtonWidth, height: moreButtonHeight)
+        
+        separator.frame = CGRect(x: 0, y: frame.height - separatorHeight, width: frame.width, height: separatorHeight)
+    }
+    
+    func setupWithEpisode(episode: Episode) {
+        
+        episodeNameLabel.text = episode.title
+        
+        let dateFormatter = DateFormatter()
```
dennisfedorko: how is this not a function on Episode? we use this in so many places

Podcast/BookmarksTableViewCell.swift
```diff
@@ -0,0 +1,228 @@
...
+        dateFormatter.dateStyle = .long
+        dateFormatter.timeStyle = .none
+        dateTimeLabel.text = dateFormatter.string(from: episode.dateCreated as Date)
```
dennisfedorko: why do you cast as Date?

Podcast/BookmarksTableViewCell.swift
```diff
@@ -0,0 +1,228 @@
...
+            if let data = try? Data(contentsOf: url) {
+                episodeImage.image = UIImage(data: data)
+            }
+        } else {
+            episodeImage.image = #imageLiteral(resourceName: "filler_image")
+        }
+    }
+    
+    // Changes numbers from 108397878 to 108.3M 
+    func convertNumberToShortenedString(n: Int) -> String {
```
dennisfedorko: should probably be an extension on Int

Podcast/BookmarksTableViewCell.swift
```diff
@@ -0,0 +1,228 @@
...
+        var final: Double = Double(n)
+        var numberOfDivisions = 0
+        while final >= 1000 {
+            final /= 1000
+            numberOfDivisions += 1
+        }
+        let suffixes = ["", "k", "M", "B", "T"]
+        var suffix = ""
```
dennisfedorko: suffix = numberOfDivisions < suffixes.count ? suffixes[numberOfDivisions] : ""

Podcast/BookmarksTableViewCell.swift
doctor3w: Can do that!

Podcast/BookmarksTableViewCell.swift
doctor3w: Not sure what this is, took it from CardTableViewCell, didn't think it was necessary but I didn't want to break anything, I'll remove it

Podcast/BookmarksTableViewCell.swift
doctor3w: I can make this into it. I think when first written in Card cell we didn't know how many times it would be used, now it's clear it's everywhere haha.

Podcast/BookmarksTableViewCell.swift
doctor3w: I'll add an Int extension so we can use this in all cells

Podcast/BookmarksTableViewCell.swift
doctor3w: v clean, me like

Podcast/BookmarkTableViewCell.swift
```diff
@@ -0,0 +1,206 @@
...
+        episodeNameLabel.textColor = .podcastBlack
+        addSubview(episodeNameLabel)
+        
+        dateTimeLabel = UILabel(frame: CGRect.zero)
+        dateTimeLabel.font = UIFont.systemFont(ofSize: 12.0)
+        dateTimeLabel.textColor = .podcastGrayDark
+        addSubview(dateTimeLabel)
+        
+        let labels: [UILabel] = [episodeNameLabel, dateTimeLabel]
+        for label in labels {
```
dennisfedorko: you don't need a for loop for this, do it for each label individually

Podcast/BookmarkTableViewCell.swift
```diff
@@ -0,0 +1,206 @@
...
+    
+    ///
+    ///Mark: Init
+    ///
+    
+    override init(style: UITableViewCellStyle, reuseIdentifier: String?) {
+        
+        super.init(style: style, reuseIdentifier: reuseIdentifier)
+        
+        frame.size.height = height
```
dennisfedorko: why are you setting the frame's height here? it's just going to be set later by the tableview

Podcast/BookmarkViewController.swift
```diff
@@ -30,20 +30,23 @@ class BookmarkViewController: UIViewController, UITableViewDelegate, UITableView
...
         bookmarkTableView.showsVerticalScrollIndicator = false
-        bookmarkTableView.register(CardTableViewCell.self, forCellReuseIdentifier: "CardTableViewCellIdentifier")
+        bookmarkTableView.register(BookmarkTableViewCell.self, forCellReuseIdentifier: "BookmarkTableViewCellIdentifier")
         view.addSubview(bookmarkTableView)
-        bookmarkTableView.rowHeight = CardTableViewCell.height
+        bookmarkTableView.rowHeight = BookmarkTableViewCell.height
         bookmarkTableView.reloadData()
         
-        cards = fetchCards()
+        episodes = fetchEpisodes()
```
dennisfedorko: you can probably assign episodes inside of the fetchEpisodes func so you don't have to return anything (especially since its going to be async)

Podcast/BookmarkTableViewCell.swift
doctor3w: Didn't write that, I'll change it

Podcast/BookmarkTableViewCell.swift
doctor3w: Didn't write that either, I'll change it

Podcast/BookmarkViewController.swift
doctor3w: Makes sense. That's what I did in my other views but I didn't write this view so I left it this way. I'll change that

PR #75, opened 2017-03-22
Faster Player

mrkbryn:
By default in iOS 10 and later, the AVPlayer stalls until it calculates that it could play until the end of the AVPlayerItem at the current download rate. However, for large podcast audio files this calculation is iffy and leads to big delays when a new track is put into the Player, even though we can safely start playing right away. This PR makes this small change to make the Player play immediately by default.


dennisfedorko:
Looks good merge it

PR #82, opened 2017-03-26
fixing jsons in models

natashaarmbrust:


Podcast/Episode.swift
```diff
@@ -13,26 +13,26 @@ class Episode: NSObject {
     
     var id: Int
     var title: String
-    var series: Series?
+    var seriesId: Int
```
dennisfedorko: seriesID should be a string

natashaarmbrust: Agreed all the ids should probably be

dennisfedorko: seriesID not seriesId

dennisfedorko:
looks good, just fix that one thing and merge

PR #83, opened 2017-03-29
Drew/better buttons

doctor3w:
Changed buttons to include text and increase hit box

### DIFF
--- a/Podcast/CardTableViewCell.swift
+++ b/Podcast/CardTableViewCell.swift
@@ -20,7 +20,8 @@ protocol CardTableViewCellDelegate: class {
 
 class CardTableViewCell: UITableViewCell {
     
-    static var height: CGFloat = 305
+    static let height: CGFloat = 305
+    static let bottomViewHeight: CGFloat = 48
     
     ///
     /// Mark: View Constants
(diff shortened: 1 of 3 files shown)


Podcast/CardTableViewCell.swift
```diff
@@ -148,11 +151,8 @@ class CardTableViewCell: UITableViewCell {
         episodeNameLabel = UILabel(frame: CGRect.zero)
         dateTimeLabel = UILabel(frame: CGRect.zero)
         descriptionLabel = UILabel(frame: CGRect.zero)
-        recommendedLabel = UILabel(frame: CGRect.zero)
-        playLabel = UILabel(frame: CGRect.zero)
-        
         
-        let labels = [episodeNameLabel, dateTimeLabel, descriptionLabel, recommendedLabel, playLabel]
+        let labels = [episodeNameLabel, dateTimeLabel, descriptionLabel]
```
dennisfedorko: the type of this should not be [UILabel?]
you shouldn't have to force unwrap in the for loop

Podcast/CardTableViewCell.swift
```diff
@@ -148,11 +151,8 @@ class CardTableViewCell: UITableViewCell {
...
         for label in labels {
             label!.textAlignment = .left
```
dennisfedorko: you should not need to force unwrap this, labels should be of type [UILabel] not [UILabel?]

dennisfedorko:
looks good, address that one comment and merge

PR #91, opened 2017-04-01
Natasha/series detial and subscriptions backend

natashaarmbrust:
massive PR sorry about that


Podcast/EndpointRequest.swift
```diff
@@ -101,7 +101,7 @@ class EndpointRequest: Operation {
         return baseURLString + path
     }
     
-    func parameters() -> [String:Any] {
+    func parameters() -> [String:Any]? {
```
dennisfedorko: why did you change this?

Podcast/EndpointRequest.swift
```diff
@@ -113,6 +113,10 @@ class EndpointRequest: Operation {
             return localQueryParameters
         }
         
+        if params.count == 0 {
```
dennisfedorko: why do you return nil instead of an empty dict?

Podcast/Episode.swift
```diff
@@ -13,26 +13,26 @@ class Episode: NSObject {
     
     var id: Int
     var title: String
-    var series: Series?
+    var seriesId: Int
```
dennisfedorko, line 16: seriesID: String

Podcast/Episode.swift
```diff
@@ -58,29 +58,47 @@ class Episode: NSObject {
...
      }
     
     // Returns data - time - series in a string
     func dateTimeSeriesString() -> String {
         let dateFormatter = DateFormatter()
         dateFormatter.dateStyle = .long
         dateFormatter.timeStyle = .none
+        let length = duration
+        
+        //add back in later when we figure out what the data is consistently
```
dennisfedorko, line 85: can you resolve this before you merge

Podcast/EpisodesBySeriesIdEndpointRequest.swift
```diff
@@ -0,0 +1,31 @@
...
+//  Copyright © 2017 Cornell App Development. All rights reserved.
+//
+
+import UIKit
+import SwiftyJSON
+
+class EpisodesBySeriesIdEndpointRequest: EndpointRequest {
+    
+    var seriesID: String
+    var offset: Int!
```
dennisfedorko: doesn't need to be Int! since you initialize in init, same with max, they can both be Int

Podcast/EpisodesBySeriesIdEndpointRequest.swift
```diff
@@ -0,0 +1,31 @@
...
+//  Podcast
+//
+//  Created by Natasha Armbrust on 3/30/17.
+//  Copyright © 2017 Cornell App Development. All rights reserved.
+//
+
+import UIKit
+import SwiftyJSON
+
+class EpisodesBySeriesIdEndpointRequest: EndpointRequest {
```
dennisfedorko: maybe FetchEpisodesForSeriesIDEndpointRequest is a better name?

Podcast/EpisodesBySeriesIdEndpointRequest.swift
```diff
@@ -0,0 +1,31 @@
...
+        self.offset = offset
+        self.max = max
+        super.init()
+        path = "/podcasts/episodes/" + seriesID
+        httpMethod = .get
+        queryParameters = ["max": max, "offset": offset]
+    }
+    
+    override func processResponseJSON(_ json: JSON) {
+        processedResponseValue = json["data"]["episodes"].map{ series in Episode(json: series.1) }
```
dennisfedorko: Im not a fan of this map format, it's not possible to understand the type of series from reading this. Also why do you use series.1?

Podcast/SearchSeriesTableViewCell.swift
```diff
@@ -81,15 +81,28 @@ class SearchSeriesTableViewCell: UITableViewCell {
...
     }
     
     func didPressSubscribeButton() {
-        subscribeButtonPressed = !subscribeButtonPressed
-        subscribeButton.isSelected = subscribeButtonPressed
-        delegate?.searchSeriesTableViewCell(cell: self, didSetSubscribeButton: subscribeButtonPressed)
+        delegate?.searchSeriesTableViewCellDidPressSubscribeButton(cell: self)
+    }
+    
+    func setSubscribeButtonToState(state: Bool) {
```
dennisfedorko: instead of state, call the parameter isSelected: Bool. makes more sense when reading if isSelected {...}

Podcast/SearchTableViewController.swift
```diff
@@ -7,6 +7,7 @@
 //
 
 import UIKit
+import NVActivityIndicatorView
```
dennisfedorko, line 10: is this the activity indicator view we're going to be using everywhere? if so I think we should make an extension on UIViewController to show the indicator view (kind of like I did with the action sheet) so you don't have to import and configure the activity indicator in every VC

Podcast/SearchTableViewController.swift
```diff
@@ -157,4 +160,32 @@ class SearchTableViewController: UITableViewController, SearchEpisodeTableViewCe
         appDelegate.showPlayer(animated: true)
         Player.sharedInstance.playEpisode(episode: episode)
     }
+    
+    func searchSeriesTableViewCellDidPressSubscribeButton(cell: SearchSeriesTableViewCell) {
+        guard let indexPath = tableView.indexPath(for:cell), let series = searchResults[.series]?[indexPath.row] as? Series else { return }
+        series.isSubscribed = !series.isSubscribed
+        if series.isSubscribed { //subscribing to series
```
dennisfedorko: I dont think your comment is adding much here

Podcast/SeriesDetailHeaderView.swift
```diff
@@ -7,9 +7,10 @@
 //
 
 import UIKit
+import NVActivityIndicatorView
```
dennisfedorko: maybe we should make it so only controllers can create activity indicators?

Podcast/SeriesDetailViewController.swift
```diff
@@ -20,76 +21,139 @@ class SeriesDetailViewController: UIViewController, SeriesDetailHeaderViewDelega
...
+        epsiodeTableView.setShouldShowInfiniteScrollHandler { _ -> Bool in
+            return self.continueInfiniteScroll
+        }
         epsiodeTableView.register(EpisodeTableViewCell.self, forCellReuseIdentifier: "EpisodeTableViewCellIdentifier")
-        
         view.addSubview(epsiodeTableView)
+    
+        epsiodeTableView.infiniteScrollIndicatorView = createLoadingAnimationView()
+        
+        if let series = self.series {
```
dennisfedorko, line 61: self.series should not be optional as that would render the entire VC useless, change the type of series to Series!

Podcast/EndpointRequest.swift
natashaarmbrust: Needs to be changed because if there are no query parameters it should return nil ... rn it's returning [:] which gave me a bug in one of the endpoint requests

Podcast/EndpointRequest.swift
natashaarmbrust: Again same reason as above

Podcast/Episode.swift
natashaarmbrust: No because I was talking to joe and he plans to standardize the data more on his end and rn some use hh:mm:ss, mm:ss, hh:mm, etc .. this will be fixed later on once we get a standard format

Podcast/EpisodesBySeriesIdEndpointRequest.swift
natashaarmbrust: The map gives you a tulle so you need the .1 .. also what would you prefer? A for loop?

Podcast/SearchTableViewController.swift
natashaarmbrust: Lol

Podcast/SeriesDetailHeaderView.swift
natashaarmbrust: It doesn't actually make one I was testing something out I'll delete that

Podcast/SeriesDetailViewController.swift
natashaarmbrust: Ok so I ran into a huge issue here with loading the table view with force unwrapping series .. the problem is we have series that can be passed in or we can make a network request for the series by the series id.. if we don't make it optional the view won't load properly

Podcast/SeriesDetailViewController.swift
```diff
@@ -20,76 +21,139 @@ class SeriesDetailViewController: UIViewController, SeriesDetailHeaderViewDelega
...
-        s.episodes = [episode]
-        series = s
-    }
-    
-    func updateViewsWithSeries(series: Series) {
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        if self.series != nil {
+            //load animation for awhile for better UX
+            let deadlineTime = DispatchTime.now() + .milliseconds(100)
```
dennisfedorko, line 79: what is the point of this function?

Podcast/SeriesDetailViewController.swift
```diff
@@ -20,76 +21,139 @@ class SeriesDetailViewController: UIViewController, SeriesDetailHeaderViewDelega
...
     }
     
-    func seriesDetailHeaderViewDidPressSubscribeButton(seriesDetailHeader: SeriesDetailHeaderView, subscribed: Bool) {
-        // Toggle subscribe (aka endpoint request)
-        
-        // When request comes back, update in case it failed
-//        seriesHeaderView.subscribeButtonChangeState(isSelected: subscribed)
+    //create and delete subscriptions
+    func seriesDetailHeaderViewDidPressSubscribeButton(seriesDetailHeader: SeriesDetailHeaderView) {
+        if !(series!.isSubscribed) { //subscribing to series
```
dennisfedorko: don't need ()

dennisfedorko: or comments

Podcast/SeriesDetailViewController.swift
```diff
@@ -129,9 +198,9 @@ class SeriesDetailViewController: UIViewController, SeriesDetailHeaderViewDelega
         return sectionHeaderHeight
     }
     
-    func tableView(_ tableView: UITableView, willDisplayHeaderView view: UIView, forSection section: Int) {
+    func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
+        let view = UIView(frame: CGRect(x: 0, y: 0, width: self.view.frame.width, height: sectionHeaderHeight))
```
dennisfedorko: rename headerView plz

Podcast/SeriesbySeriesIdEndpointRequest.swift
```diff
@@ -0,0 +1,26 @@
...
+//  Podcast
+//
+//  Created by Natasha Armbrust on 3/30/17.
+//  Copyright © 2017 Cornell App Development. All rights reserved.
+//
+
+import UIKit
+import SwiftyJSON
+
+class SeriesBySeriesIdEndpointRequest: EndpointRequest {
```
dennisfedorko: FetchSeriesForSeriesIDEndpointRequest

Podcast/TabbedPageViewController.swift
```diff
@@ -251,9 +256,26 @@ class TabbedPageViewController: UIViewController, UIPageViewControllerDataSource
...
+        case .series:
+            guard let series = searchResults[.series]?[index] as? Series else { return }
+            searchResultsDelegate?.didTapOnSeriesCell(series: series)
+        case .people:
+            //present external profile view here
+            let dummyViewController = UIViewController()
+            presentingViewController?.navigationController?.pushViewController(dummyViewController, animated: true)
+        case .tags:
+            //present tag view here
+            let dummyViewController = UIViewController()
```
dennisfedorko: didn't you make the tag vc already?

Podcast/SeriesDetailViewController.swift
natashaarmbrust: oh okay so i thought UX perspective it looked a little weird to have the series just show up .. so now what happens is the loader animates for a second then series header view shows up. It looks better this way

Podcast/UserSubscriptionsEndpointRequest.swift
```diff
@@ -0,0 +1,27 @@
...
+//  Podcast
+//
+//  Created by Natasha Armbrust on 3/29/17.
+//  Copyright © 2017 Cornell App Development. All rights reserved.
+//
+
+import UIKit
+import SwiftyJSON
+
+class UserSubscriptionsEndpointRequest: EndpointRequest {
```
dennisfedorko: FetchUserSubscriptionsEndpointRequest

Podcast/TabbedPageViewController.swift
natashaarmbrust: oh yea i can put that in

Podcast/UserSubscriptionsEndpointRequest.swift
```diff
@@ -0,0 +1,27 @@
...
+    init(userID: String) {
+        self.userID = userID
+        super.init()
+        path = "/subscriptions"
+        httpMethod = .get
+        queryParameters = ["id": userID]
+    }
+    
+    override func processResponseJSON(_ json: JSON) {
+        processedResponseValue = json["data"]["subscriptions"].map{ series in SubscriptionSeries(json: series.1) }
```
dennisfedorko: clean up map by making the type for series explicit

natashaarmbrust: not quite sure what you mean by this

Podcast/EndpointRequest.swift
dennisfedorko: what bug did it cause? this goes straight into the alamofire request initializer, so it shouldn't have any effect if its empty

Podcast/Episode.swift
dennisfedorko: ok, mark with a //TODO

Podcast/EpisodesBySeriesIdEndpointRequest.swift
dennisfedorko: why would mapping json give you a tuple? if you call some variable series, it should be assumed that its of type Series.

Podcast/SeriesDetailViewController.swift
dennisfedorko: then clearly something is wrong with the setup of the controller. in what case would we pass it in and in which case can we make a network request?

Podcast/SeriesDetailViewController.swift
dennisfedorko: ok ill take your word for it

Podcast/EndpointRequest.swift
natashaarmbrust: It times out on the request cuz it's malformed

Podcast/EpisodesBySeriesIdEndpointRequest.swift
natashaarmbrust: Addressed

Podcast/Episode.swift
natashaarmbrust: Other PR included this

Podcast/Episode.swift
natashaarmbrust: Addressed

Podcast/EpisodesBySeriesIdEndpointRequest.swift
natashaarmbrust: Addressed

Podcast/SearchSeriesTableViewCell.swift
natashaarmbrust: Addressed

Podcast/SeriesDetailViewController.swift
natashaarmbrust: Ok so in search we have full series we are passing in. Since we already get full series from search we don't need another network request before the view is presented but only need one to get the episodes. However in subscriptions since our data is denormalized on backend we only get part of a series so I make a network request to get the full series and show the loading indicator while I do that

Podcast/SeriesDetailViewController.swift
natashaarmbrust: Addressed

Podcast/SeriesDetailViewController.swift
natashaarmbrust: Addressed

Podcast/UserSubscriptionsEndpointRequest.swift
natashaarmbrust: Addressed

Podcast/EndpointRequest.swift
natashaarmbrust: What happens is you call parameters() in the request which ends up being [:] if there are none. However alamofire is expecting nil if there are node

Podcast/EpisodesBySeriesIdEndpointRequest.swift
natashaarmbrust: it returns (string, json) tuple indicating which number value it is in the json array ... i think its fine i'm not sure what else you'd like me to use

Podcast/SeriesDetailViewController.swift
dennisfedorko: I think we discussed this

Issue #85, opened 2017-03-29
Create deep links for podcasts for sharing


natashaarmbrust:
For MVP1, let's just reference the sharing to the podcast app and not a deeplink.. In the future, sharing an episode should reference that specific episode with a deeplink

dennisfedorko:
what does this mean?

doctor3w:
Deep links are fairly simple. We just need a URL format for our app. Otherwise I'm not sure how we'll share anything.

PR #94, opened 2017-04-10
add share sheet to action sheet and episode detail view

kevingreer:
#77 

I've added the share sheets, but you can't actually share anything yet (I think that's where #85 comes in)


dennisfedorko:
Looks good, merge it

PR #96, opened 2017-04-12
Episode Detail

mrkbryn:
Adds episode detail views in the places where it should appear. Seeing as the designs are likely to change, especially related to the comments, I didn't want to waste too much time on adding every little detail if it may get removed later. I can polish things up in the next iteration.

### DIFF
--- a/Podcast/Comment.swift
+++ b/Podcast/Comment.swift
@@ -1,0 +1,25 @@
+//
+//  Comment.swift
+//  Podcast
+//
+//  Created by Mark Bryan on 4/12/17.
+//  Copyright © 2017 Cornell App Development. All rights reserved.
+//
+
+import UIKit
+
+class Comment: NSObject {
+    var episodeId: String
+    var creator: String
+    var text: String
+    var creationDate: String
+    var time: String
+    
+    init(episodeId: String, creator: String, text: String, creationDate: String, time: String) {
+        self.episodeId = episodeId
+        self.creator = creator
+        self.text = text
+        self.creationDate = creationDate
+        self.time = time
+    }
+}
--- a/Podcast/Comment.swift
+++ b/Podcast/Comment.swift
@@ -8,7 +8,7 @@
 
 import UIKit
 
-class Comment: NSObject {
+class Comment {
     var episodeId: String
     var creator: String
     var text: String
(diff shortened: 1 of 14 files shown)


Podcast/Comment.swift
```diff
@@ -0,0 +1,25 @@
...
+//  Comment.swift
+//  Podcast
+//
+//  Created by Mark Bryan on 4/12/17.
+//  Copyright © 2017 Cornell App Development. All rights reserved.
+//
+
+import UIKit
+
+class Comment: NSObject {
```
dennisfedorko: you don't need to subclass NSObject

Podcast/CommentsTableViewCell.swift
```diff
@@ -0,0 +1,100 @@
...
+import UIKit
+
+protocol CommentsTableViewCellDelegate {
+    func commentCellDidPressLikeButton(cell: CommentsTableViewCell)
+    func commentCellDidPressRepliesButton(cell: CommentsTableViewCell)
+}
+
+class CommentsTableViewCell: UITableViewCell {
+    let minimumHeight: CGFloat = 200
+    let imageViewDimension: CGFloat = 30
```
dennisfedorko: imageViewSize and make it a CGSize would probably make more sense

dennisfedorko: or imageViewDiameter

Podcast/Episode.swift
```diff
@@ -100,5 +100,12 @@ class Episode: NSObject {
         // Check if series title is empty because some are
         return seriesTitle != "" ? "\(dateFormatter.string(from: dateCreated)) • \(length) • \(seriesTitle)" : "\(dateFormatter.string(from: dateCreated)) • \(length)"
     }
+    
+    func dateString() -> String {
```
dennisfedorko, line 104: don't we have another function that does this?

Podcast/PlayButton.swift
```diff
@@ -10,8 +10,12 @@ import UIKit
 
 class PlayButton: UIButton {
     
-    init() {
-        super.init(frame: .zero)
+    convenience init() {
```
dennisfedorko: what is the point of this when all it does is just do the same thing as initializing with UIButton()

Podcast/Episode.swift
mrkbryn: The other function we have returns "data - time - series", but in this case we just need "date."

dennisfedorko: ok forget this

PR #99, opened 2017-04-22
HTML Labels and Improved Episode Detail

mrkbryn:
Connects Episode Detail to everything (I'm pretty sure I got every place that it should come up..)
Also uses HTML attributed text for labels so the descriptions look better in episode detail.


natashaarmbrust:
A couple things, can you get the delegate protocol working for the episode header ... i.e. when you press heart icon, bookmark, and more icon. (also you are using the wrong bookmark icon .. you should be using the skinnier version)

In addition, can you connect up the cells to listening history! (its in profile tab)

Podcast/EpisodeDetailViewController.swift
```diff
@@ -8,51 +8,99 @@
...
 
-class EpisodeDetailViewController: UIViewController, UITableViewDataSource, UITableViewDelegate {
-    var scrollView: UIScrollView!
-    var episodeDetailHeaderView: EpisodeDetailHeaderView!
+class EpisodeDetailViewController: UITableViewController, EpisodeDetailHeaderViewCellDelegate {
     var episode: Episode?
-    var commentsTableViewHeader: CommentsTableViewHeader!
-    var commentsTableView: UITableView!
     var comments: [Comment] = Array.init(repeating: Comment(episodeId: "", creator: "Mark Bryan", text: "Great point here!!", creationDate: "5 months ago", time: "5:13"), count: 10)
+    var episodeDetailViewHeaderHeight: CGFloat?
```
dennisfedorko: why is this optional?

Podcast/EpisodeDetailViewController.swift
```diff
@@ -8,51 +8,99 @@
...
     }
     
-    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        guard let cell = tableView.dequeueReusableCell(withIdentifier: "CommentsTableViewCellIdentifier") as? CommentsTableViewCell else { return UITableViewCell() }
-        cell.setupForComment(comment: comments[indexPath.row])
-        commentsTableView.rowHeight = cell.frame.height
-        return cell
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        if indexPath.section == 0 {
+            if let cell = tableView.dequeueReusableCell(withIdentifier: "EpisodeDetailHeaderViewCellIdentifier") as? EpisodeDetailHeaderViewCell, let episode = episode {
```
dennisfedorko: you can combine these two if statements into one

Podcast/EpisodeDetailViewController.swift
```diff
@@ -8,51 +8,99 @@
...
+                cell.setupForEpisode(episode: episode)
+                cell.delegate = self
+                cell.layoutSubviews()
+                episodeDetailViewHeaderHeight = cell.frame.height
+                return cell
+            }
+        } else {
+            if let cell = tableView.dequeueReusableCell(withIdentifier: "CommentsTableViewCellIdentifier") as? CommentsTableViewCell {
```
dennisfedorko: you can combine this with the else

Podcast/EpisodeDetailViewController.swift
```diff
@@ -8,51 +8,99 @@
...
+                cell.setupForComment(comment: comments[indexPath.row])
+                return cell
+            }
+        }
+        return UITableViewCell()
     }
     
-    func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+    override func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        if section == 0 {
```
dennisfedorko: return section == 0 ? 1 : comments.count

Podcast/EpisodeDetailViewController.swift
mrkbryn: because it's set at run time once the cell has setup its subviews. i could also just set it to 0 and overwrite that value once the height is determined. that's probably better

PR #104, opened 2017-04-23
Parts of Profile Backend

doctor3w:
User and subscriptions and following and unfollowing should work. Still needs favorites and following is not represented in the user until Joe fixes backend.

### DIFF
--- a/Podcast/ExternalProfileViewController.swift
+++ b/Podcast/ExternalProfileViewController.swift
@@ -7,6 +7,7 @@
 //
 
 import UIKit
+import NVActivityIndicatorView
 
 class ExternalProfileViewController: UIViewController, UITableViewDataSource, UITableViewDelegate, ProfileHeaderViewDelegate, RecommendedSeriesTableViewCellDelegate, RecommendedSeriesTableViewCellDataSource, RecommendedEpisodesOuterTableViewCellDelegate, RecommendedEpisodesOuterTableViewCellDataSource {
     
@@ -16,6 +17,8 @@ class ExternalProfileViewController: UIViewController, UITableViewDataSource, UI
     var miniHeader: ProfileMiniHeader!
     var profileTableView: UITableView!
     var backButton: UIButton!
+    
+    var loadingAnimation: NVActivityIndicatorView!
     
     let headerViewHeight = ProfileHeaderView.height
     let miniBarHeight = ProfileHeaderView.miniBarHeight
@@ -33,15 +36,11 @@ class ExternalProfileViewController: UIViewController, UITableViewDataSource, UI
     
     var user: User!
     
+    var favorites: [Episode]!
+    var subscriptions: [GridSeries]!
+    
     override func viewDidLoad() {
         super.viewDidLoad()
-        createSubviews()
-        fetchUser(id: 0)
-    }
-    
-    func createSubviews() {
-        guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else { return }
-        
         view.backgroundColor = .podcastWhiteDark
         
         backButton = UIButton(type: .custom)
@@ -49,6 +48,18 @@ class ExternalProfileViewController: UIViewController, UITableViewDataSource, UI
         backButton.setImage(UIImage(named: "backArrow"), for: .normal)
         backButton.adjustsImageWhenHighlighted = true
         backButton.addTarget(self, action: #selector(didPressBackButton), for: .touchUpInside)
+        view.addSubview(backButton)
+        view.bringSubview(toFront: backButton)
+        
+        loadingAnimation = createLoadingAnimationView()
+        loadingAnimation.center = view.center
+        view.addSubview(loadingAnimation)
+        loadingAnimation.startAnimating()
+        UIApplication.shared.statusBarStyle = .default
+    }
+    
+    func createSubviews() {
+        guard let appDelegate = UIApplication.shared.delegate as? AppDelegate else { return }
         
         // Instantiate tableView
         profileTableView = UITableView(frame: CGRect(x: 0, y: 0, width: view.frame.width, height: view.frame.height), style: .grouped)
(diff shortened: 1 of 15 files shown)


Podcast/ExternalProfileViewController.swift
```diff
@@ -93,47 +104,96 @@ class ExternalProfileViewController: UIViewController, UITableViewDataSource, UI
...
+    func fetchUser(id: String) {
         
-        // Dummy data
-        let user = User()
-        user.firstName = "Paul"
-        user.lastName = "Dugg"
-        user.username = "doglover12"
-        user.numberOfFollowing = 100
-        user.numberOfFollowers = 4
+        let userRequest = UserByIDEndpointRequest(userID: id)
```
dennisfedorko: change the name of this network request to something like FetchUserByIDEndpointRequest

Podcast/ExternalProfileViewController.swift
```diff
@@ -93,47 +104,96 @@ class ExternalProfileViewController: UIViewController, UITableViewDataSource, UI
...
+                    UIApplication.shared.statusBarStyle = .lightContent
+                }
+            }
+        }
         
-        let s = Series()
-        s.title = "Design Details"
-        s.numberOfSubscribers = 832567
-        user.subscriptions = Array(repeating: s, count: 7)
+        System.endpointRequestQueue.addOperation(userRequest)
```
dennisfedorko: what happens if this fails?

Podcast/ExternalProfileViewController.swift
```diff
@@ -93,47 +104,96 @@ class ExternalProfileViewController: UIViewController, UITableViewDataSource, UI
...
         self.user = user
         // Update views
         profileHeaderView.setUser(user)
         miniHeader.setUser(user)
-        profileTableView.reloadData()
     }
     
     func didPressBackButton() {
-        let _ = navigationController?.popViewController(animated: true)
+        if let navController = navigationController {
```
dennisfedorko: navigationController?.popViewController(animated: true)

Podcast/ExternalProfileViewController.swift
```diff
@@ -93,47 +104,96 @@ class ExternalProfileViewController: UIViewController, UITableViewDataSource, UI
...
+            navController.popViewController(animated: true)
+        }
     }
     
     // Mark: - ProfileHeaderView
     func profileHeaderDidPressFollowButton(profileHeader: ProfileHeaderView, follow: Bool) {
         // Follow/Unfollow someone
+        if follow {
+            let newFollowRequest = NewFollowingEndpointRequest(userID: user.id)
```
dennisfedorko: should probably rename endpoint request to FollowUserEndpointRequest

Podcast/ExternalProfileViewController.swift
```diff
@@ -93,47 +104,96 @@ class ExternalProfileViewController: UIViewController, UITableViewDataSource, UI
...
+                }
+            }
+            newFollowRequest.failure = { (endpointRequest: EndpointRequest) in
+                DispatchQueue.main.async {
+                    // Follow failed
+                    profileHeader.followButton.isSelected = false
+                    self.user.isFollowing = false
+                }
+            }
+            System.endpointRequestQueue.addOperation(newFollowRequest)
```
dennisfedorko: you should probably have something to prevent user from tapping follow/unfollow multiple times

Podcast/ExternalProfileViewController.swift
```diff
@@ -93,47 +104,96 @@ class ExternalProfileViewController: UIViewController, UITableViewDataSource, UI
...
-        let _ = navigationController?.popViewController(animated: true)
+        if let navController = navigationController {
+            navController.popViewController(animated: true)
+        }
     }
     
     // Mark: - ProfileHeaderView
     func profileHeaderDidPressFollowButton(profileHeader: ProfileHeaderView, follow: Bool) {
         // Follow/Unfollow someone
+        if follow {
```
dennisfedorko: split this up into two helper functions, one for following, one for unfollowing

Podcast/ExternalProfileViewController.swift
```diff
@@ -209,28 +270,35 @@ class ExternalProfileViewController: UIViewController, UITableViewDataSource, UI
...
     
     func recommendedEpisodesTableViewCell(cell: RecommendedEpisodesOuterTableViewCell, dataForItemAt indexPath: IndexPath) -> Episode {
-        return user.favoriteEpisodes[indexPath.row]
+        guard let favoriteEpisodes = favorites else { return Episode() }
+        return favoriteEpisodes[indexPath.row]
     }
     
     func numberOfRecommendedEpisodes(forRecommendedEpisodesOuterTableViewCell cell: RecommendedEpisodesOuterTableViewCell) -> Int {
-        return user.favoriteEpisodes.count
+        return 0
```
dennisfedorko, line 325: you return 0 and then do stuff...

Podcast/InternalProfileViewController.swift
```diff
@@ -45,11 +45,22 @@ class InternalProfileViewController: UIViewController, UITableViewDelegate, UITa
         view.addSubview(tableView)
     }
     
+    override func viewDidAppear(_ animated: Bool) {
```
dennisfedorko: delete

Podcast/InternalProfileViewController.swift
```diff
@@ -45,11 +45,22 @@ class InternalProfileViewController: UIViewController, UITableViewDelegate, UITa
...
+//        navigationItem.backBarButtonItem = nil
+    }
+    
     // MARK: InternalProfileHeaderViewDelegate
     
     func internalProfileHeaderViewDidPressViewProfile(internalProfileHeaderView: InternalProfileHeaderView) {
         let myProfileViewController = ExternalProfileViewController()
-        myProfileViewController.user = System.currentUser
+//        let navigationVC = UINavigationController(rootViewController: myProfileViewController)
```
dennisfedorko: delete commented out code

Podcast/InternalProfileViewController.swift
```diff
@@ -45,11 +45,22 @@ class InternalProfileViewController: UIViewController, UITableViewDelegate, UITa
...
+//        navigationItem.backBarButtonItem = UIBarButtonItem(title: "", style: .plain, target: nil, action: nil)
+        if let currentUser = System.currentUser {
+            myProfileViewController.fetchUser(id: currentUser.id)
+//            myProfileViewController.fetchUser(id: "9d417b19-2474-11e7-be6b-3f00bd823d20") // Test for other people 
+        } else {
+            myProfileViewController.fetchUser(id: "")
```
dennisfedorko: what does this do?

Podcast/TagViewController.swift
```diff
@@ -13,7 +13,7 @@ class TagViewController: UIViewController, UITableViewDelegate, UITableViewDataS
     var tableView: UITableView!
     var tag: Tag!
     var episodes: [Episode] = []
-    var series: [Series] = []
+    var series: [SubscriptionSeries] = []
```
dennisfedorko: SubscriptionSeries is kind of confusing, can we come up with a better name?

Podcast/ExternalProfileViewController.swift
doctor3w: That gives a warning when I did that, so Natasha changed it I believe. What's the problem with this way?

Podcast/ExternalProfileViewController.swift
doctor3w: Yeah we don't have favorites yet so this is temporary

Podcast/InternalProfileViewController.swift
doctor3w: I wasn't actually sure what to do in this case, what do you recommend?

Podcast/TagViewController.swift
doctor3w: Oh yeah I was supposed to change it to GridSeries

Podcast/ExternalProfileViewController.swift
dennisfedorko: let _ = navigationController?.popViewController(animated: true) should not give a warning

Podcast/ExternalProfileViewController.swift
dennisfedorko: oh ok

Podcast/InternalProfileViewController.swift
dennisfedorko: there's no case where you make it to this point and you don't have a current user, so do a guard let instead of an if let and just return in the else condition

Podcast/ExternalProfileViewController.swift
doctor3w: I could disable it until the request has completed

PR #106, opened 2017-04-28
Buttons

mrkbryn:
- Creates custom UIButtons so that we are using the same buttons across all views. This reduces a lot of duplicate code and allows us to change the look of a certain type of button without having to change >20 lines of code in many different files
- Fixes layout issue with `PlayButton` in the episode search cells. The cell now readjusts the layout of the subviews when the `PlayButton` is changed to be "Play" or "Playing"

In addition to created custom UIButton classes, I think we should expand this to the other common views we use across many different ViewControllers. For example, we could easily wrap the bottom button panel with the separator line into its own view. We use this in a ton of different views and it'd be easier to keep this view consistent (i.e. layout constraints and dimensions) across all ViewControllers if we just write the layout code once in a custom UIView class.

### DIFF
--- a/Podcast/BookmarkTableViewCell.swift
+++ b/Podcast/BookmarkTableViewCell.swift
@@ -9,7 +9,6 @@
 import UIKit
 
 protocol BookmarkTableViewCellDelegate: class {
-    
     func bookmarkTableViewCellDidPressPlayPauseButton(bookmarksTableViewCell: BookmarkTableViewCell)
     func bookmarkTableViewCellDidPressRecommendButton(bookmarksTableViewCell: BookmarkTableViewCell)
     func bookmarkTableViewCellDidPressMoreActionsButton(bookmarksTableViewCell: BookmarkTableViewCell)
@@ -31,8 +30,6 @@ class BookmarkTableViewCell: UITableViewCell {
     let dateTimeLabelX: CGFloat = 90
     let dateTimeLabelY: CGFloat = 41.5
     let dateTimeLabelHeight: CGFloat = 14.5
-    
-    let buttonTitlePadding: CGFloat = 7
     
     let playButtonX: CGFloat = 90
     let playButtonBottomY: CGFloat = 18
@@ -57,9 +54,9 @@ class BookmarkTableViewCell: UITableViewCell {
     var episodeImage: UIImageView!
     var episodeNameLabel: UILabel!
     var dateTimeLabel: UILabel!
-    var recommendedButton: UIButton!
-    var moreButton: UIButton!
-    var playButton: UIButton!
+    var recommendedButton: RecommendButton!
+    var moreButton: MoreButton!
+    var playButton: PlayButton!
     var separator: UIView!
     
     weak var delegate: BookmarkTableViewCellDelegate?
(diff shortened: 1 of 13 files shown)


mrkbryn:
Addresses #103

Podcast/BookmarkTableViewCell.swift
```diff
@@ -171,7 +171,7 @@ class BookmarkTableViewCell: UITableViewCell {
     }
     
     func setPlayButtonToState(isPlaying: Bool) {
-        playButton.isEnabled = !isPlaying
+        playButton.isSelected = isPlaying
```
doctor3w, line 174: Wait I made this disabled while playing on purpose because there was a bug when you held it down.

mrkbryn: I don't think we want to use `isEnabled` as that determines if you can click on a UIButton. I see what you're talking about with that bug, though, and I just realized that I forgot to change `isEnabled` to `isSelected` in a couple places. I'm about to commit a change that fixes it for all views. I think some people were using `isEnabled` and others were using `isSelected` which is what was causing the bug.

mrkbryn: Just checked and that bug is fixed with my latest commit. You can hold down the play button as long as you want and it'll play and update the UI correctly if you release from inside the button

Podcast/EpisodeTableViewCell.swift
```diff
@@ -160,41 +157,24 @@ class EpisodeTableViewCell: UITableViewCell {
         podcastImage = UIImageView(frame: CGRect.zero)
         mainView.addSubview(podcastImage)
         
-        bookmarkButton = UIButton(frame: CGRect.zero)
-        bookmarkButton.setImage(#imageLiteral(resourceName: "bookmark_feed_icon_unselected"), for: .normal)
-        bookmarkButton.setImage(#imageLiteral(resourceName: "bookmark_feed_icon_selected"), for: .selected)
-        bookmarkButton.addTarget(self, action: #selector(didPressBookmarkButton), for: .touchUpInside)
-        bottomView.addSubview(bookmarkButton)
+        // buttons
```
dennisfedorko: useless comment

Podcast/EpisodeTableViewCell.swift
```diff
@@ -160,41 +157,24 @@ class EpisodeTableViewCell: UITableViewCell {
...
-        playButton.contentHorizontalAlignment = .left
-        playButton.titleEdgeInsets = UIEdgeInsets(top: 0, left: buttonTitlePadding, bottom: 0, right: 0)
-        playButton.titleLabel?.font = .systemFont(ofSize: 12)
+        
+        bottomView.addSubview(bookmarkButton)
+        bottomView.addSubview(recommendedButton)
+        bottomView.addSubview(moreButton)
         bottomView.addSubview(playButton)
+
+//        playButton.setTitle("Play", for: .normal)
```
dennisfedorko: delete

Podcast/FeedControlButton.swift
```diff
@@ -0,0 +1,22 @@
...
+//  FeedControlButton.swift
+//  Podcast
+//
+//  Created by Mark Bryan on 4/27/17.
+//  Copyright © 2017 Cornell App Development. All rights reserved.
+//
+
+import UIKit
+
+class FeedControlButton: UIButton {
```
dennisfedorko, line 11: what does this button do? unclear from name

dennisfedorko:
Looks good, address comments and merge

PR #107, opened 2017-04-28
Fixed tab bar height with accessory view

doctor3w:
Fixes the issue where tableView insets don't increase when the mini player is showing (or not showing).


dennisfedorko:
looks good merge it

PR #112, opened 2017-05-07
Kevin/episode detail

kevingreer:
#110

### DIFF
--- a/Podcast/EpisodeDetailHeaderViewCell.swift
+++ b/Podcast/EpisodeDetailHeaderViewCell.swift
@@ -172,7 +172,6 @@ class EpisodeDetailHeaderViewCell: UITableViewCell {
         episodeTitleLabel.text = episode.title
         dateLabel.text = episode.dateString()
         descriptionLabel.attributedText = episode.attributedDescriptionString()
-        descriptionLabel.text = "hihdiahwdlajh alkjwhd  lakjd alkjw d lkajwhdal lakjhdalwkjhd alkjwhd  dlkajhwdljkaw lkjhdalkwjd  kdjhaw k jhd alkjwhd lkjawhd ljdhawlkjdh lkjd hakwjhd lkj lkjdhalkjwhdlkjh wlk j dakjhwl jkh lkjh"
     }
     
     func setPlaying(playing: Bool) {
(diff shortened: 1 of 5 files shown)


Podcast/EpisodeDetailHeaderViewCell.swift
```diff
@@ -94,7 +95,8 @@ class EpisodeDetailHeaderViewCell: UITableViewCell {
         addSubview(episodeTitleLabel)
         
         dateLabel = UILabel(frame: CGRect(x: marginSpacing, y: 0, width: frame.width - 2 * marginSpacing, height: 0))
-        dateLabel.font = UIFont.systemFont(ofSize: 14)
+        dateLabel.font = UIFont.systemFont(ofSize: 12)
+        dateLabel.textColor = UIColor.colorFromCode(0x959ba5)
```
natashaarmbrust: put this in utilities! it might already be in there .. if not make it named podcast[ColorDescription]

Podcast/SearchPeopleTableViewCell.swift
```diff
@@ -79,7 +79,11 @@ class SearchPeopleTableViewCell: UITableViewCell {
     
     func configure(for user: User, index: Int) {
         self.index = index
-        profilePictureImageView.image = #imageLiteral(resourceName: "sample_profile_pic")
+        if let url = user.imageURL, let data = try? Data(contentsOf: url) {
```
natashaarmbrust: profilePictureImageView to be ImageView then call setLoadImageUrlAsync(url:...) or wtvr the function is called.

Podcast/SearchPeopleTableViewCell.swift
```diff
@@ -79,7 +79,11 @@ class SearchPeopleTableViewCell: UITableViewCell {
...
+            profilePictureImageView.image = UIImage(data: data)
+        } else {
+            profilePictureImageView.image = #imageLiteral(resourceName: "sample_profile_pic")
```
natashaarmbrust: Also why are you not doing UIImage(name: "sample_profile_pic")? If there is not a good reason then switch to that

kevingreer: what's the difference?

natashaarmbrust: the first one is what we use throughout the app.. we should keep it consistent

PR #122, opened 2017-07-22
Update pods, remove unused ones, add SwiftLint

dantheli:
This removes old pods that are no longer used, many of which were required by Facebook Login. This also adds SwiftLint. A future PR will include addition of our SwiftLint config file.


natashaarmbrust:
LGTM

PR #144, opened 2017-09-12
Natasha/episode button utility bar

natashaarmbrust:
Fixing issue #145 
- Fixing episode detail delegate methods
- Adding episode utility button bar (refactor because used in both episode cells and episode detail view)
- TODO: add it to episode table view cell as well (maybe refactor card and episode table view cells so we don't have to keep adding things to both)

Two images below now have refactoring EpisodeUtilityButtonBarView 

screen shot 2017-09-12 at 2 03 32 am
screen shot 2017-09-12 at 2 03 38 am


natashaarmbrust:
@drewsdunne Please Review

Podcast/CardTableViewCell.swift
```diff
@@ -81,19 +61,14 @@ class CardTableViewCell: UITableViewCell {
...
-    var lineSeperator: UIView!
     var topLineSeperator: UIView!
-    var moreButton: MoreButton!
-    var playButton: PlayButton!
     var contextLabel: UILabel!
     var contextImages: [ImageView] = []
     var contextView: UIView! //view for upper context bar of feed cell
     var mainView: UIView! //main view
-    var bottomView: UIView! //bottom bar view with buttons
+    var utilityButtonBarView: EpisodeUtilityButtonBarView! //bottom bar view with buttons
```
doctor3w, line 71: Why not call this episodeButtonBar, utilityButtonBarView might not be clear initially what it contains other than it is reused buttons.

doctor3w (commented):
Looks good. Def wanna bring this out to episode cell. Maybe even could refactor more since the whole episode part is the same. Or make individual parts and then one big reusable part with the little parts.

Podcast/CardTableViewCell.swift
natashaarmbrust: Fair I can change it to episodeUtilityButtonBarView

Issue #140, opened 2017-09-06
Profile view - Series icons subtitle doesn't fit


doctor3w:
The "Last updated j..." looks awful. Why does our prewritten text not even fit. 

![image](https://user-images.githubusercontent.com/1222887/30139546-8d70d924-933c-11e7-8f8f-238cf3969c1d.png)

Issue #128, opened 2017-08-31
Episode buttons on series detail do not work 


natashaarmbrust:
<img width="371" alt="screen shot 2017-08-30 at 9 55 30 pm" src="https://user-images.githubusercontent.com/14966713/29902749-023a6762-8dce-11e7-92ed-d5946231471d.png">

Like, bookmark, and more button on series detail view episode cell are all disabled

PR #151, opened 2017-09-14
Natasha/series fixes

natashaarmbrust:
- Makes Series subclass of GridSeries
- Fixes texts in GridSeries collection cells  #140
- Series detail buttons now work #128
screen shot 2017-09-14 at 1 47 36 am

screen shot 2017-09-14 at 1 47 29 am


natashaarmbrust:
@drewsdunne Please Review

Podcast/DiscoverViewController.swift
```diff
@@ -64,7 +64,7 @@ class DiscoverViewController: ViewController, UITableViewDelegate, UITableViewDa
         
         // Populate with dummy data
         let s = GridSeries()
-        s.seriesTitle = "Design Details"
+        s.title = "Design Details"
```
doctor3w, line 67: Omg thank you, I hated this before

Podcast/SeriesType.swift
```diff
@@ -0,0 +1,14 @@
...
+//  SeriesType.swift
+//  Podcast
+//
+//  Created by Natasha Armbrust on 9/14/17.
+//  Copyright © 2017 Cornell App Development. All rights reserved.
+//
+
+import UIKit
+
+enum SeriesType {
```
doctor3w: Do we really need a whole enum for this?

doctor3w (commented):
Looks great, just had one question.

Podcast/SeriesType.swift
natashaarmbrust: No we don't actually and we shouldn't because now one is a subclass of the other so we can just use casting. I'll remove this.

PR #162, opened 2017-09-27
Kevin/python backend

kevingreer:


Podcast/FeedElement.swift
```diff
@@ -0,0 +1,53 @@
...
+//
+//  Created by Kevin Greer on 9/27/17.
+//  Copyright © 2017 Cornell App Development. All rights reserved.
+//
+
+import Foundation
+import SwiftyJSON
+
+enum FeedContext: String {
+    case followingRecommendation = "FOLLOWING_RECOMMENDATION"
```
natashaarmbrust, line 13: hmm thinking these could be more descriptive ... such as:
episodeRecommendationFromUser
seriesSubscriptionFromUser
episodeReleaseFromSeries

Podcast/FeedElement.swift
```diff
@@ -0,0 +1,53 @@
...
+    case followingSubscription = "FOLLOWING_SUBSCRIPTION"
+    case newlyReleasedEpisode = "NEW_SUBSCRIBED_EPISODE"
+}
+
+class FeedElement: NSObject {
+    
+    var context: FeedContext
```
natashaarmbrust: comment these!!!

Podcast/GridSeries.swift
```diff
@@ -17,21 +17,19 @@ class GridSeries: NSObject {
     var largeArtworkImageURL: URL?
     var smallArtworkImageURL: URL?
     var isSubscribed: Bool
-    var lastUpdated: Date
```
natashaarmbrust: Why are we removing lastUpdated from a GridSeries? This is what we use for cells in Subscription View (which has text of the last updated episode)

Podcast/Series.swift
```diff
@@ -19,31 +19,28 @@ class Series: GridSeries {
...
-        let numberOfSubscribers = json["n_subscribers"].intValue
-        let tags = json["genres"].arrayValue.map({ (tag: JSON) in Tag(name: tag.stringValue) })
-        let lastUpdatedString = json["last_updated"].stringValue
-        let lastUpdated = DateFormatter.parsingDateFormatter.date(from: lastUpdatedString) ?? Date()
+        let numberOfSubscribers = json["subscribers_count"].intValue
+        let tags = json["genres"].stringValue.components(separatedBy: ";").map({ tag in Tag(name: tag)})
         let smallArtworkURL = URL(string: json["image_url_sm"].stringValue)
         let largeArtworkURL = URL(string: json["image_url_lg"].stringValue)
-        self.init(id: id, title: title, author: author, descriptionText: descriptionText, smallArtworkImageURL: smallArtworkURL, largeArtworkImageURL: largeArtworkURL, tags: tags, numberOfSubscribers: numberOfSubscribers, isSubscribed: isSubscribed, lastUpdated: lastUpdated)
+        self.init(id: id, title: title, author: author, descriptionText: descriptionText, smallArtworkImageURL: smallArtworkURL, largeArtworkImageURL: largeArtworkURL, tags: tags, numberOfSubscribers: numberOfSubscribers, isSubscribed: isSubscribed)
```
natashaarmbrust: Again with lastUpdated comment

Podcast/FetchFeedEndpointRequest.swift
```diff
@@ -11,34 +11,22 @@ import SwiftyJSON
 
 class FetchFeedEndpointRequest: EndpointRequest {
     
-    var offset: Int
-    var max: Int
+    var pageSize: Int
```
natashaarmbrust, line 14: This will give us errors in FeedViewController. We need to also be fixing this View Controller before we push this.

Podcast/GridSeries.swift
natashaarmbrust: If we removed lastUpdated from here we will also need to change SeriesGridCollectionViewCell to reflect these changes (rn it is currently giving an error ... for now we can comment out and stub in a TODO)

natashaarmbrust (changes requested):

natashaarmbrust (changes requested):
Commented on some issues

Podcast/FeedElement.swift
kevingreer: I think these are actually more descriptive, if we decide to add recommendations from really popular users that you might not follow then that would also be an episodeRecommendationFromUser but not a followingRecommendation. I'll change it if you'd still prefer your suggestions though

PR #166, opened 2017-10-06
adding autolayout to grid series

natashaarmbrust:
- Changed SeriesGridCollectionViewCell to use autolayout .. this fixes loading issues that frame based layout created


Podcast/BookmarkTableViewCell.swift
```diff
@@ -148,12 +148,9 @@ class BookmarkTableViewCell: UITableViewCell {
         recommendedButton.setTitle(numberOfRecommendations, for: .normal)
         recommendedButton.setTitle(numberOfRecommendations, for: .selected)
         recommendedButton.isSelected = episode.isRecommended
-        
-        episodeImage.image = #imageLiteral(resourceName: "nullSeries")
+    
+        episodeImage.setImageAsynchronouslyWithDefaultImage(url: episode.smallArtworkImageURL)
```
dantheli: Who wrote ImageView.swift? Not sure why we don't just use Haneke (or a better library that isn't outdated af) directly

Podcast/ImageView.swift
```diff
@@ -12,6 +12,14 @@ class ImageView: UIImageView {
         }
     }
     
+    func setImageAsynchronouslyWithDefaultImage(url: URL?, defualtImage: UIImage = #imageLiteral(resourceName: "nullSeries")) {
```
dantheli: 'defualt' lol
But more importanlty was there a reason to use this? You're not using `defualtImage` anywhere

I really am not sure what this class is for, especially since Haneke has the ability to set default image and completion callback and we aren't using either

dantheli (changes requested):

Podcast/BookmarkTableViewCell.swift
natashaarmbrust: It's a nice abstraction to have a wrapper class so not every file has to import Haneke

Podcast/ImageView.swift
natashaarmbrust: I do use the defaultImage override in a lot of places. I removed the completion callback

Podcast/ImageView.swift
```diff
@@ -3,23 +3,7 @@ import Haneke
...
-    }
-    
-    func setImageAsynchronously(url: URL, completion: (() -> ())?) {
-        if frame == .zero {
-            print("Error: Cannot set image asynchronously with an ImageView with frame .zero")
-            return
-        }
-        setImageCompletionBlock = completion
-        hnk_setImage(from: url)
+    func setImageAsynchronouslyWithDefaultImage(url: URL?, defualtImage: UIImage = #imageLiteral(resourceName: "nullSeries")) {
```
dantheli: Fix typo and GTM

PR #167, opened 2017-10-07
finishing new styleguide

natashaarmbrust:


mindylou (approved):
I'm a fan of the new color names, looks good!

PR #181, opened 2017-10-24
Refactor Dynamic Settings pages

doctor3w:
Created singleton UserSettings object for use around the app. These settings can then also be added to the settings page as well fairly easily.


Podcast/SettingsPageViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+        super.viewDidLoad()
+        navigationController?.setNavigationBarHidden(false, animated: false)
+        
+        tableView = UITableView(frame: CGRect.zero, style: .grouped)
+        tableView.showsVerticalScrollIndicator = true
+        tableView.allowsSelection = false
+        tableView.backgroundColor = .paleGrey
+        for type in SettingsFieldType.allValues {
+            let (reuseId, cellClass):(String, AnyClass) = type.reuseIdAndClass
+            tableView.register(cellClass, forCellReuseIdentifier: reuseId)
```
mindylou: This is so nice!!

Podcast/SettingsTableViewCell.swift
```diff
@@ -12,18 +12,25 @@ class SettingsTableViewCell: UITableViewCell {
...
         titleLabel.font = ._14RegularFont()
         titleLabel.textColor = .offBlack
         titleLabel.textAlignment = .left
         addSubview(titleLabel)
+        
+        titleLabel.snp.makeConstraints({ make in
+            make.right.lessThanOrEqualToSuperview().inset(labelSidePadding)
+            make.left.equalToSuperview().inset(labelSidePadding)
+            make.top.equalToSuperview().inset(labelTopPadding)
+            make.bottom.equalToSuperview().inset(labelTopPadding)
```
mindylou: you can chain some of these for clarity like `make.top.bottom.equaltoSuperview().insert(labelTopPadding)`

Podcast/UserSettings.swift
```diff
@@ -0,0 +1,76 @@
...
+                            System.currentUser = resp["user"] as? User
+                        }
+                        changeUsernameVC.navigationController?.popViewController(animated: true)
+                    }
+                    changeUsernameEndpointRequest.failure = { (endpointRequest: EndpointRequest) in
+                        //changeUsernameVC.setError(at: IndexPath(row: 0, section: 0), with: "Bad Username")
+                        // TODO: add error handling
+                    }
+                    System.endpointRequestQueue.addOperation(changeUsernameEndpointRequest)
+                }, type: .textField, placeholder: "new username")
```
mindylou, line 68: It might make more sense for the placeholder to be the current username, not sure if designs have come in for this yet. Also this placeholder text doesn't show up for me:
image

Podcast/UserSettings.swift
```diff
@@ -0,0 +1,76 @@
...
+    
+    static let sharedSettings = UserSettings()
+    
+    private var saveCounter: Int!
+    private let maxSave: Int = 5
+    
+    var pushNotifications: Bool!
+    
+    private override init() {
+        saveCounter = 0
```
mindylou: What is this for? I don't see it being used anywhere

mindylou (changes requested):
This looks good! Just one issue with the placeholder text for the settings with text fields

Podcast/SettingsPageViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+// Add more field types as created
+// String value is cell reuse ID
+enum SettingsFieldType {
+    case button
+    case textField
+    case disclosure
+    case label
+    case toggle
+    
+    static let allValues: [SettingsFieldType] = [.button, .textField, .disclosure, .label, .toggle]
```
natashaarmbrust, line 20: are you sure there isn't a field on an enum likes SettingsFieldType.values that is already created for us?

Podcast/SettingsPageViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+    
+    var reuseIdAndClass: (String, AnyClass) {
```
natashaarmbrust: let's put the reuseIdentifier on the tableViewCell as a field so then we can just have a tableViewCell variable on this enum 
I think the tuple return with this unspecific field name might get confusing later on

Podcast/SettingsPageViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+            } else {
+                navigationItem.rightBarButtonItem = nil
+            }
+        }
+    }
+    var returnOnSave:Bool = true
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        navigationController?.setNavigationBarHidden(false, animated: false)
```
natashaarmbrust, line 119: is this not the default? or do we need to do this because we are coming from internal profile?

Podcast/SettingsPageViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+        
+        tableView = UITableView(frame: CGRect.zero, style: .grouped)
+        tableView.showsVerticalScrollIndicator = true
+        tableView.allowsSelection = false
+        tableView.backgroundColor = .paleGrey
+        for type in SettingsFieldType.allValues {
+            let (reuseId, cellClass):(String, AnyClass) = type.reuseIdAndClass
+            tableView.register(cellClass, forCellReuseIdentifier: reuseId)
+        }
+        tableView.separatorInset = UIEdgeInsetsMake(0, 0, 0, 0)
```
natashaarmbrust, line 128: Do we need this? isn't the default to have no edge insets

Podcast/SettingsPageViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+        tableView.dataSource = self
+        tableView.delegate = self
+        tableView.allowsSelection = true
```
natashaarmbrust, line 131: just above you had tableView.allowsSelection = false .. not sure which one we want but delete one of them

Podcast/SettingsPageViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+        mainScrollView = tableView
+        view.addSubview(tableView)
+        
+        saveButton = UIBarButtonItem(barButtonSystemItem: .save, target: self, action: #selector(saveAction))
+        if showSave {
```
natashaarmbrust, line 136: isn't this in the didSet of showSave ... I am confused here.

Podcast/SettingsPageViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+            navigationItem.rightBarButtonItem = saveButton
+        } else {
+            navigationItem.rightBarButtonItem = nil
+        }
+        
+        tableView.snp.makeConstraints({ make in
```
natashaarmbrust: YAY! snapkit

natashaarmbrust: you don't need the parenthesis tho!

natashaarmbrust: makeConstraints { make in 
....
}

Podcast/SettingsPageViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+        super.viewWillAppear(animated)
+        tableView.reloadData()
+    }
+    
+    @objc private func saveAction() {
+        for (i, section) in sections.enumerated() {
+            for (j, setting) in section.items.enumerated() {
+                switch setting.type {
+                case .textField:
+                    guard let cell = tableView.cellForRow(at: IndexPath(row: j, section: i)) as? TextFieldSettingsTableViewCell else {
```
natashaarmbrust, line 157: very nice

Podcast/SettingsPageViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+                }
+            }
+        }
+        if returnOnSave {
+            self.navigationController?.popViewController(animated: true)
+        }
+    }
+    
+    func setError(at index: IndexPath, with message: String) {
+        // TODO: need fixing
```
natashaarmbrust, line 172: what fixing? like display a pop-up?

Podcast/SettingsPageViewController.swift
```diff
@@ -0,0 +1,207 @@
...
+        cell.accessoryType = .none
+        switch setting.type {
+        case .disclosure:
+            cell.accessoryType = .disclosureIndicator
+            break
+        case .button:
+            cell.titleLabel.textColor = .blue
+            break
+        case .textField:
+            let tcell = cell as? TextFieldSettingsTableViewCell ?? TextFieldSettingsTableViewCell()
```
natashaarmbrust, line 205: should this maybe be a guard let ?

Podcast/SettingsTableViewCell.swift
```diff
@@ -32,6 +39,15 @@ class SettingsTableViewCell: UITableViewCell {
     
     func setTitle(_ title: String) {
         titleLabel.text = title
+        titleLabel.sizeToFit()
```
natashaarmbrust, line 41: Autolayout should do a sizeToFit automatically if constraints are correctly set

Podcast/TextFieldSettingsTableViewCell.swift
```diff
@@ -0,0 +1,71 @@
...
+        
+        textField = UITextField(frame: CGRect.zero)
+        textField.font = UIFont._14RegularFont()
+        textField.textColor = .offBlack
+        textField.autocapitalizationType = .none
+        textField.autocorrectionType = .no
+        addSubview(textField)
+
+        textField.snp.makeConstraints({ make in
+            make.left.equalTo(titleLabel.snp.right).offset(textFieldPadding)
```
natashaarmbrust: Always do leading and trailing instead of left and right because languages that read right to left can then use our app too!

Podcast/SettingsTableViewCell.swift
```diff
@@ -12,18 +12,25 @@ class SettingsTableViewCell: UITableViewCell {
...
         super.init(style: style, reuseIdentifier: reuseIdentifier)
-        titleLabel = UILabel(frame: CGRect(x: labelPadding, y: (height - labelHeight) / 2, width: frame.width - 2 * labelPadding, height: labelHeight))
+        titleLabel = UILabel(frame: CGRect.zero)
         titleLabel.font = ._14RegularFont()
         titleLabel.textColor = .offBlack
         titleLabel.textAlignment = .left
         addSubview(titleLabel)
+        
+        titleLabel.snp.makeConstraints({ make in
+            make.right.lessThanOrEqualToSuperview().inset(labelSidePadding)
```
natashaarmbrust: leading and trailing instead of left to right!

Podcast/TextFieldSettingsTableViewCell.swift
```diff
@@ -0,0 +1,71 @@
...
+            // no insets top and bottom (easier to tap)
+            make.top.equalToSuperview()
+            make.bottom.equalToSuperview()
+            make.width.greaterThanOrEqualTo(textFieldWidth)
+        })
+        
+        accessoryView = UIImageView(image: #imageLiteral(resourceName: "failure_icon"))
+        addSubview(accessoryView!)
+        accessoryView?.snp.makeConstraints({ make in
+            make.left.equalTo(textField.snp.right)
```
natashaarmbrust: are you sure we need all these constraints? also leading and trailing !

Podcast/UserSettings.swift
```diff
@@ -0,0 +1,76 @@
...
+//
+
+import UIKit
+
+class UserSettings: NSObject {
+    
+    static let sharedSettings = UserSettings()
+    
+    private var saveCounter: Int!
+    private let maxSave: Int = 5
```
natashaarmbrust, line 16: ?

Podcast/UserSettings.swift
```diff
@@ -0,0 +1,76 @@
...
+    
+    private func loadData() {
+        pushNotifications = UserDefaults.standard.bool(forKey: "push_notifications")
+    }
+    
+    func writeToFile() {
+        UserDefaults.standard.set(pushNotifications, forKey: "push_notifications")
+    }
+    
+    static let changeUsernamePage: SettingsPageViewController = {
```
natashaarmbrust: let's comment these two static fields a bit because its getting a bit confusing how all these components are interacting together

natashaarmbrust (changes requested):

Podcast/SettingsTableViewCell.swift
doctor3w: Lol I should learn more snapkit

Podcast/UserSettings.swift
doctor3w: It will be haha when I finish. It's going to be much more nice in the future. Plan is to use it to determine when to periodically save to UserDefaults because UserDefaults is really slow.

Podcast/UserSettings.swift
doctor3w: Yeah I had a plan for that, I'm rethinking some stuff so this will change soon.

Podcast/SettingsPageViewController.swift
doctor3w: Not according to stack overflow lmao

Podcast/SettingsPageViewController.swift
doctor3w: Yeah that sounds good!

Podcast/SettingsPageViewController.swift
doctor3w: Yeah it's necessary because of some weird stuff in profile. I've been looking into a way to fix this.

Podcast/SettingsPageViewController.swift
doctor3w: No the default I believe has like a left of 15 I think...

Podcast/SettingsPageViewController.swift
doctor3w: lmao yeah I want true...kind of...

Podcast/SettingsPageViewController.swift
doctor3w: So I had it in the didSet, and it didn't work...so I put it here. Guess I never went back to figuring out this. I'll look into it again.

Podcast/SettingsPageViewController.swift
doctor3w: It auto makes parentheses though haha.

Podcast/SettingsPageViewController.swift
doctor3w: My accessory view wasn't working but I decided to wait on fixing this until I asked eileen about the design for what should happen anyway.

Podcast/SettingsPageViewController.swift
doctor3w: Yeah this shouldn't happen, but maybe.

Podcast/SettingsTableViewCell.swift
doctor3w: But do I want this flipped for languages that are right to left instead?

Podcast/UserSettings.swift
doctor3w: You'll see. It'll be really cool.

Podcast/UserSettings.swift
doctor3w: And okay will do. I'm working on a better way to define these.

Podcast/SettingsPageViewController.swift
doctor3w: Actually can't do that, classes can't override static variables, and I don't want to instantiate an entire class just to get a string value out.

Podcast/SettingsPageViewController.swift
doctor3w: Ahh you need it in both, navigationItem doesn't exist the first time you set it.

natashaarmbrust (approved):
LGTM, I am not able to run it because of the backend error rn but I trust it works / before you merge in make sure it works.

mindylou (approved):

PR #182, opened 2017-10-26
fixing profile bug

natashaarmbrust:
- We made all the followers / following have current user as the input... But we can look at another users followers / following .. this PR accounts for this


mindylou (approved):
looks good!

PR #185, opened 2017-11-01
Mindy/search itunes

### LINKED ISSUE #176: Search iTunes!
screen shot 2017-10-22 at 2 57 54 pm
This design is missing the people tab, but we should search iTunes for the series tab (series tab only!)

mindylou:
- implement search iTunes feature for series (tap on header to search query in iTunes)
- closes #176 

- currently backend only fetches a certain amount so infinite scrolling isn't working (repeats the same cells)


Podcast/SearchTableViewController.swift
```diff
@@ -93,6 +103,89 @@ class SearchTableViewController: ViewController, UITableViewDelegate, UITableVie
         currentlyPlayingIndexPath = nil
     }
     
+    func setupSearchITunesHeader() {
```
natashaarmbrust: Let's abstract this away to a file in Views called SearchItunesView

Podcast/TapGestureRecognizer+.swift
```diff
@@ -0,0 +1,46 @@
...
+//  TapGestureRecognizer+.swift
+//  Podcast
+//
+//  Created by Mindy Lou on 10/30/17.
+//  Copyright © 2017 Cornell App Development. All rights reserved.
+//
+
+import UIKit
+
+extension UITapGestureRecognizer {
```
natashaarmbrust: Where do we use this?

Podcast/SearchTableViewController.swift
```diff
@@ -13,6 +13,7 @@ enum SearchType {
     case episodes
     case series
     case people
+    case itunes
```
natashaarmbrust: I get this error when clicking (search Itunes directly)
screen shot 2017-11-01 at 11 35 38 am
screen shot 2017-11-01 at 11 35 26 am

Podcast/SearchTableViewController.swift
```diff
@@ -23,6 +24,8 @@ enum SearchType {
             return "Series"
         case .people:
             return "People"
+        case .itunes:
```
natashaarmbrust: When trying to subscribe to a series 
screen shot 2017-11-01 at 11 38 55 am

natashaarmbrust (changes requested):
Good work and it's almost there, but I think we should change the design on this. I would revert all the changes to TabbedViewController and SearchTableViewController except adding the iTunes banner header (and being able to dismiss it). I would make a new ViewController SearchItunesViewController that includes a search bar and a search results table view within that view controller (the cells would be SearchSeriesTableVIewCells). This controller should pop up from the bottom (much like the action sheet) when the "search iTunes directly" banner is tapped. The controller should also be easily dismissed via a cancel button at the top. 

Reasons for this approach: adding the logic of iTunes into the SearchTableViewController is getting confusing / there are a lot of edge cases we need to handle with indexes. We plan to search iTunes as only a beta feature (then maybe hide it behind settings in the future) so removing it entirely as a separate search will help us do that later on. Finally, the UI treatment is a bit confusing right now. I can't tell if I am searching iTunes or just our app. I think the solution a proposed should fix this and also be easier for you!

Podcast/TapGestureRecognizer+.swift
mindylou: i was trying to get it to recognize tapping on a subsection of attributed text (like a link) in a uilabel, but it's not working for multi-line labels. i can remove it for now

Podcast/PlayerViewController.swift
```diff
@@ -194,16 +194,14 @@ class PlayerViewController: TabBarAccessoryViewController, PlayerDelegate, Playe
...
-            switch rate {
-            case .normal:
-                Player.sharedInstance.setSpeed(rate: .fast)
-            case .fast:
-                Player.sharedInstance.setSpeed(rate: .slow)
-            case .slow:
-                Player.sharedInstance.setSpeed(rate: .normal)
-            }
+        let rate = Player.sharedInstance.getSpeed()
+        switch rate {
```
natashaarmbrust: put in a comment for this. I can tell what it does but its a bit confusing lol.

Podcast/SearchITunesTableViewController.swift
```diff
@@ -30,10 +30,11 @@ class SearchITunesTableViewController: ViewController, UITableViewDelegate, UITa
         mainScrollView = tableView
         
         loadingIndicatorView = createLoadingAnimationView()
-        loadingIndicatorView!.center = CGPoint(x: UIScreen.main.bounds.width / 2, y: UIScreen.main.bounds.height / 2)
         view.addSubview(loadingIndicatorView!)
+        loadingIndicatorView?.snp.makeConstraints { make in
```
natashaarmbrust: why is this loadingIndicator an optional? EmptyStateTableView now has a loading indicator within it so we don't need this loading indicator.

Podcast/SearchITunesTableViewController.swift
```diff
@@ -100,6 +101,11 @@ class SearchITunesTableViewController: ViewController, UITableViewDelegate, UITa
         fetchData(query: searchText, completion: nil)
     }
     
+    func searchBarCancelButtonClicked(_ searchBar: UISearchBar) {
+        tableView.isHidden = false
+        tableView.backgroundView?.isHidden = true
```
natashaarmbrust: reason for hiding the background view? I guess this is what drew was saying with the two different background views for our table views.

Podcast/SearchITunesTableViewController.swift
```diff
@@ -111,30 +117,7 @@ class SearchITunesTableViewController: ViewController, UITableViewDelegate, UITa
...
-                series.didUnsubscribe()
-                cell.setSubscribeButtonToState(isSubscribed: series.isSubscribed)
-            }
-            deleteSubscriptionEndpointRequest.failure = { (endpointRequest: EndpointRequest) in
-                series.isSubscribed = true
-                cell.setSubscribeButtonToState(isSubscribed: series.isSubscribed)
-            }
-            System.endpointRequestQueue.addOperation(deleteSubscriptionEndpointRequest)
-        }
+        series.subscriptionChange(completion: cell.setSubscribeButtonToState(isSubscribed:numberOfSubscribers:))
```
natashaarmbrust: you can just do series.subscriptionChange(completion: cell.setSubscribeButtonToState)

Podcast/SearchITunesViewController.swift
```diff
@@ -16,7 +16,7 @@ class SearchITunesViewController: ViewController {
     
     override func viewDidLoad() {
         super.viewDidLoad()
-        view.backgroundColor = .offWhite
+        view.backgroundColor = .paleGrey
         title = "Search iTunes"
         
         searchResultsController = SearchITunesTableViewController()
```
natashaarmbrust: is there a reason you did separate SearchITunesTableViewController and SearchITunesViewController? If not, we should actually combine these. You can say `searchResultsController = self` and make this view controller have a tableView within it

Podcast/SearchTableViewController.swift
```diff
@@ -73,9 +72,11 @@ class SearchTableViewController: ViewController, UITableViewDelegate, UITableVie
         tableView.delegate = self
         tableView.dataSource = self
         tableView.infiniteScrollIndicatorView = createLoadingAnimationView()
+        tableView.tableFooterView = UIView()
```
natashaarmbrust: reason for this?

Podcast/SearchTableViewController.swift
```diff
@@ -161,6 +162,7 @@ class SearchTableViewController: ViewController, UITableViewDelegate, UITableVie
     
     class func buildListOfAllSearchTableViewControllerTypes() -> [SearchTableViewController] {
         let searchTableViewControllerEpisodes = SearchTableViewController()
+        searchTableViewControllerEpisodes.updateTableViewInsetsForAccessoryView()
```
natashaarmbrust: reason for this?

Podcast/ViewController.swift
```diff
@@ -29,6 +29,7 @@ class ViewController: UIViewController {
         guard let appDelegate = UIApplication.shared.delegate as? AppDelegate, let mainScrollView = mainScrollView else { return }
         if appDelegate.tabBarController.accessoryViewController == nil {
             mainScrollView.contentInset.bottom = appDelegate.tabBarController.tabBarHeight
+            mainScrollView.scrollIndicatorInsets = UIEdgeInsetsMake(0, 0, 0, appDelegate.tabBarController.tabBarHeight)
```
natashaarmbrust: Does this work? If not, let's uncommit this file.

natashaarmbrust:
One more thing, let's make the dismiss button on the search itunes header view be larger or at least cover a larger span so that people will acutally be able to press it!

natashaarmbrust:
screen shot 2017-11-04 at 5 25 50 pm
What is this weird button on the side of search?

Podcast/SearchITunesTableViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+    
+    func fetchData(query: String, completion: (() -> ())?)  {
+        System.endpointRequestQueue.cancelAllEndpointRequestsOfType(type: SearchEpisodesEndpointRequest.self)
+        System.endpointRequestQueue.cancelAllEndpointRequestsOfType(type: SearchUsersEndpointRequest.self)
+        System.endpointRequestQueue.cancelAllEndpointRequestsOfType(type: SearchSeriesEndpointRequest.self)
+        
+        tableView.backgroundView?.isHidden = true
+        loadingIndicatorView?.startAnimating()
+        
+        let request = SearchITunesEndpointRequest(query: query, offset: 0, max: 20)
```
natashaarmbrust: is it pagenated now? if so we can add a moving offset

Podcast/SearchITunesTableViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+        
+        loadingIndicatorView = createLoadingAnimationView()
+        view.addSubview(loadingIndicatorView!)
+        loadingIndicatorView?.snp.makeConstraints { make in
+            make.center.equalToSuperview()
+        }
+        
+    }
+    
+    func fetchData(query: String, completion: (() -> ())?)  {
```
natashaarmbrust: do we use this completion? I don't see the point of it. Let's remove it

Podcast/SearchITunesTableViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+        System.endpointRequestQueue.cancelAllEndpointRequestsOfType(type: SearchEpisodesEndpointRequest.self)
+        System.endpointRequestQueue.cancelAllEndpointRequestsOfType(type: SearchUsersEndpointRequest.self)
+        System.endpointRequestQueue.cancelAllEndpointRequestsOfType(type: SearchSeriesEndpointRequest.self)
+
```
natashaarmbrust: The UI for the view is a bit weird when you try to search multiple times. Let's actually clear the searchResults when we make a request and reload the tableview because right now the loading indicator pops up over the cells and it looks weird.

Podcast/SearchITunesTableViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+    func numberOfSections(in tableView: UITableView) -> Int {
+        return 1
+    }
+    
+    func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+        return seriesCellHeight
+    }
+    
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        if let cell = tableView.dequeueReusableCell(withIdentifier: seriesCellIdentifier) as? SearchSeriesTableViewCell {
```
natashaarmbrust: if let cell = ....  as? SearchSeriesTableViewCell else { cell = SearchSeriesTableViewCell() }

then remove the duplicated code

Podcast/SearchITunesTableViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+    func updateSearchResults(for searchController: UISearchController) {
+        tableView.backgroundView?.isHidden = true
+    }
+    
+    // MARK: SearchSeriesCellDelegate
+    
+    func searchSeriesTableViewCellDidPressSubscribeButton(cell: SearchSeriesTableViewCell) {
+        guard let indexPath = tableView.indexPath(for: cell) else { return }
+        let series = searchResults[indexPath.row]
+        series.subscriptionChange(completion: cell.setSubscribeButtonToState(isSubscribed:numberOfSubscribers:))
```
natashaarmbrust: same here with the `series.subscriptionChange(completion: cell.setSubscribeButtonToState)`

Podcast/SearchITunesViewController.swift
```diff
@@ -0,0 +1,65 @@
...
+        searchController.searchBar.placeholder = "Search"
+        searchController.searchBar.delegate = searchResultsController
+        searchController.delegate = searchResultsController
+        searchController.hidesNavigationBarDuringPresentation = false
+        searchController.isActive = true
+        searchController.obscuresBackgroundDuringPresentation = false
+        searchController.dimsBackgroundDuringPresentation = false
+        definesPresentationContext = true
+
+        if #available(iOS 11.0, *) {
```
doctor3w: We're only on iOS 11 so we shouldn't need this.

natashaarmbrust:
screen shot 2017-11-04 at 5 53 17 pm
going to a series detail still doesn't work because navigation controller is nil. I believe this is because of your setup with the SearchITunesViewController and SearchITunesTableViewController. When you combine these into one make sure this now works!

Podcast/ViewController.swift
doctor3w: Hmm yeah I don't think this is what we want.

doctor3w (changes requested):

Podcast/SearchTableViewController.swift
mindylou: a lot of the stuff i did in searchtableviewcontroller was to see if i could figure out the inset bug, so i was trying to add a footerview to bump the tableview upwards past the tab bar

Podcast/ViewController.swift
mindylou: lol nope it doesn't, i'll undo that

Podcast/SearchITunesTableViewController.swift
mindylou: nope, according to joe there's no way to paginate itunes search results

Podcast/SearchITunesViewController.swift
```diff
@@ -0,0 +1,134 @@
...
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        view.backgroundColor = .paleGrey
+        title = "Search iTunes"
+        
+        tableView = EmptyStateTableView(frame: view.frame, type: .search)
+        tableView.dataSource = self
+        tableView.delegate = self
+        tableView.stopLoadingAnimation()
+        tableView.backgroundView?.isHidden = true
```
natashaarmbrust, line 30: isn't background view hidden the default for emptystatetableview

Podcast/SearchITunesViewController.swift
```diff
@@ -0,0 +1,134 @@
...
+        request.success = { request in
+            guard let results = request.processedResponseValue as? [Series] else { return }
+            if results.isEmpty {
+                self.tableView.backgroundView?.isHidden = false
+            }
+            self.searchResults = results
+            self.tableView.stopLoadingAnimation()
+            self.tableView.reloadData()
+        }
+        request.failure = { request in
```
natashaarmbrust: you never use request so you can do `{ _ in` instead of `{ request in`

Podcast/SearchITunesViewController.swift
mindylou: yeah it is, but for some reason isHidden is false for this tableView

PR #187, opened 2017-11-01
Simple background audio

doctor3w:
This took a while to figure out but we need AudioToolbox and AVFoundation. They weren't actually added to our project. 

We also definitely don't want Mix With Others. It sounded hilarious with music playing over a podcast. 

I probably need to remove the observer, but I'm not sure where would be the proper place to do that other than termination. 

Also, if the audio is interrupted when playing (aka it's paused), when coming back into the app the play button looks as if it was never paused (still playing). I'm pretty sure it's because while it's in the background, the delegate UI update calls aren't made (I put a print in there and it stopped every time it went into the background, probably because the main thread isn't running anymore).

### DIFF
--- a/Podcast/AppDelegate.swift
+++ b/Podcast/AppDelegate.swift
@@ -81,14 +81,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         loginNavigationController.setNavigationBarHidden(true, animated: false)
         
         // AVAudioSession
-        var audioOptions: AVAudioSessionCategoryOptions = [.mixWithOthers]
-        if #available(iOS 10, *) {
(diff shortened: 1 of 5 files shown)


Podcast/AppDelegate.swift
```diff
@@ -88,6 +99,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         return true
     }
     
+    @objc func beginInterruption() {
+        Player.sharedInstance.pause()
+        print("interrupted")
```
natashaarmbrust, line 104: Is this the print statement you are referring too? Maybe explain more in the meeting since I am a bit confused

natashaarmbrust (approved):
LGTM!

PR #199, opened 2017-11-06
Fix all known broken constraints

dantheli:
All known constraints. There may be broken constraints in episode cells.

### DIFF
--- a/Podcast/SeriesDetailViewController.swift
+++ b/Podcast/SeriesDetailViewController.swift
@@ -38,17 +38,11 @@ class SeriesDetailViewController: ViewController, SeriesDetailHeaderViewDelegate
     override func viewDidLoad() {
         super.viewDidLoad()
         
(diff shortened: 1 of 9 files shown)


Podcast/SeriesDetailViewController.swift
```diff
@@ -38,17 +38,11 @@ class SeriesDetailViewController: ViewController, SeriesDetailHeaderViewDelegate
     override func viewDidLoad() {
         super.viewDidLoad()
         
-        seriesHeaderView = SeriesDetailHeaderView(frame: CGRect(x: 0, y: 0, width: view.frame.width, height: seriesHeaderViewMinHeight))
+        seriesHeaderView = SeriesDetailHeaderView(frame: CGRect(x: 0.0, y: 0.0, width: view.frame.height, height: seriesHeaderViewMinHeight))
```
natashaarmbrust: why would we want the header view width to be view.frame.height?

natashaarmbrust (approved):
LGTM just that one line I am confused about.

Podcast/SeriesDetailViewController.swift
dantheli: oops

PR #207, opened 2017-11-10
Natasha/bugs

natashaarmbrust:
- Fixes `EpisodeDetailViewController` setup including removing comments, adding series title as a button, and changing the header cell setup to be a view 
- Fixes pull to refresh bug of the refresh control being on top of the tableview
- Let's TagCollectionViewCells be auto-resizing and hopefully fixes this issue 
- Fixes date parsing from an episode which caused episodes to all be the same time
- Fixes `SeriesDetailViewController` from `ExternalProfileViewController` isSubscribed bug


Podcast/EmptyStateTableView.swift
```diff
@@ -71,6 +77,24 @@ class EmptyStateTableView: UITableView, EmptyStateViewDelegate {
...
     }
     
+    //MARK
+    //MARK: Refresh Control Functions
+    //MARK
+    
+    func endRefreshing() {
+        if let control = refreshControl {
+            if control.isRefreshing {
+                DispatchTime.waitFor(milliseconds: 500, completion: { self.refreshControl?.endRefreshing() } )
```
mindylou, line 87: why is there a delay of 500 ms?

Podcast/EpisodeDetailHeaderViewCell.swift
```diff
@@ -88,45 +73,50 @@ class EpisodeDetailHeaderViewCell: UITableViewCell {
...
-        seriesTitleLabel.frame = CGRect(x: seriesTitleLabelXValue,y: marginSpacing, width: frame.width - marginSpacing - seriesTitleLabelXValue, height: 0)
-        UILabel.adjustHeightToFit(label: seriesTitleLabel, numberOfLines: 2)
         
-        publisherLabel.frame = CGRect(x: publisherLabelXValue,
-                                      y: seriesTitleLabel.frame.maxY + marginSpacing / 2,
-                                      width: frame.width - marginSpacing - publisherLabelXValue,
-                                      height: 0)
-        UILabel.adjustHeightToFit(label: publisherLabel)
+        episodeArtworkImageView.snp.makeConstraints { make in
+            make.size.equalTo(artworkDimension)
```
mindylou: why do you set the size in frame-based before this and then use snapkit?

Podcast/EpisodeDetailViewController.swift
```diff
@@ -8,100 +8,84 @@
...
+        view.backgroundColor = .offWhite
+        
+        episodeDescriptionView.isEditable = false
+        episodeDescriptionView.font = ._14RegularFont()
+        episodeDescriptionView.textColor = .charcoalGrey
+        episodeDescriptionView.showsVerticalScrollIndicator = false
+        episodeDescriptionView.backgroundColor = .clear
+        episodeDescriptionView.contentInset = UIEdgeInsetsMake(marginSpacing / 2, marginSpacing, marginSpacing, marginSpacing)
+        view.addSubview(episodeDescriptionView)
+        mainScrollView = episodeDescriptionView
```
mindylou: image
the description scrollview starts in the middle of the episode description

Podcast/FollowerFollowingViewController.swift
```diff
@@ -76,10 +72,14 @@ class FollowerFollowingViewController: ViewController, UITableViewDataSource, UI
...
             guard let follows = request.processedResponseValue as? [User] else { return }
             self.users = follows
-            self.refreshControl.endRefreshing()
+            self.usersTableView.endRefreshing()
             self.usersTableView.stopLoadingAnimation()
             self.usersTableView.reloadSections([0] , with: .automatic)
         }
+        
+        endpointRequest.failure = { _ in
+            self.usersTableView.endRefreshing()
```
mindylou, line 79: i got this error in the console, but it kept refreshing?
```
Could not load user subscriptions, request failed
Could not load user favorites, request failed
```

mindylou (changes requested):
some small changes but looks good!

Podcast/EpisodeDetailViewController.swift
mindylou: from the profile, tapping on a recommended episode does this:
image

Podcast/EmptyStateTableView.swift
natashaarmbrust: We need this for the refresher not too be on top of the table view during refresh. It's what solves the bug. The problem is, when we refresh views that result in no changes it returns too fast.

Podcast/EpisodeDetailHeaderViewCell.swift
natashaarmbrust: Because ImageView can't be initialized to zero frame or else Haneke throughs an error.

Podcast/EpisodeDetailViewController.swift
natashaarmbrust: This is weird. Are you still on iPhone 7 small screen? What was the iphone you used to get this bug?

mindylou: it was on the iphone 7 simulator

Podcast/FollowerFollowingViewController.swift
natashaarmbrust: These print statements are from ExternalProfileViewController not FollowerFollowingViewController so the endpoint probably hadn't returned yet in your case.
screen shot 2017-11-10 at 4 55 19 pm

mindylou: ohhhhh ok!

mindylou (approved):

Podcast/EmptyStateTableView.swift
```diff
@@ -23,7 +23,8 @@ class EmptyStateTableView: UITableView, EmptyStateViewDelegate {
     var loadingAnimation: NVActivityIndicatorView!
     weak var emptyStateTableViewDelegate: EmptyStateTableViewDelegate?
     
-    init(frame: CGRect, type: EmptyStateType) {
+    // isRefreshable Boolean indicating whether this tableView has a UIRefreshControl
+    init(frame: CGRect, type: EmptyStateType, isRefreshable: Bool = false) {
```
doctor3w, line 27: nice

Podcast/Episode.swift
```diff
@@ -128,14 +108,15 @@ class Episode: NSObject {
         return dateFormatter.string(from: dateCreated)
     }
     
-    func attributedDescriptionString() -> NSAttributedString {
+    func attributedDescriptionString() -> NSMutableAttributedString {
```
doctor3w: Why return a mutable one? Just use a mutable one and return a regular one? Idk if this is actually more efficient but it doesn't seem like it should be mutable after being returned.

Podcast/EpisodeDetailViewController.swift
```diff
@@ -8,100 +8,95 @@
...
-    
-    var tableView: UITableView!
+class EpisodeDetailViewController: ViewController, EpisodeDetailHeaderViewDelegate {
+
+    let marginSpacing: CGFloat = EpisodeDetailHeaderView.marginSpacing
     var episode: Episode?
-    var comments: [Comment] = Array.init(repeating: Comment(episodeId: "", creator: "Mark Bryan", text: "Great point here!!", creationDate: "5 months ago", time: "5:13"), count: 10)
-    var episodeDetailViewHeaderHeight: CGFloat = 0
+    var headerView: EpisodeDetailHeaderView = EpisodeDetailHeaderView()
+    var episodeDescriptionView: UITextView = UITextView()
```
doctor3w, line 16: I'm not sure a text view is what we want. If a description is too long the whole view (including the header) should scroll. So either the header and text view need to be put into a scroll view, or keep the table view logic.

Podcast/ExternalProfileViewController.swift
```diff
@@ -309,7 +309,9 @@ class ExternalProfileViewController: ViewController, UITableViewDataSource, UITa
     }
     
     func recommendedSeriesTableViewCell(cell: RecommendedSeriesTableViewCell, didSelectItemAt indexPath: IndexPath) {
-        let seriesDetailViewController = SeriesDetailViewController(series: subscriptions[indexPath.row])
+        let seriesDetailViewController = SeriesDetailViewController()
+        // fetch new series because isSubscribed is different for current user
+        seriesDetailViewController.fetchSeries(seriesID: subscriptions[indexPath.row].seriesId)
```
doctor3w, line 314: Hmm we shouldn't need to fetch it all over again. The original fetch should have this data already.

doctor3w: And if not it needs to be added, we shouldn't be setting this locally on fetch.

Podcast/PlayerEpisodeDetailView.swift
```diff
@@ -89,7 +89,7 @@ class PlayerEpisodeDetailView: UIView {
         episodeArtworkImageView.setImageAsynchronouslyWithDefaultImage(url: episode.largeArtworkImageURL)
         episodeTitleLabel.text = episode.title
         dateLabel.text = episode.dateTimeSeriesString()
-        let mutableString = NSMutableAttributedString(attributedString: episode.attributedDescriptionString())
+        let mutableString = episode.attributedDescriptionString()
         mutableString.addAttribute(NSAttributedStringKey.foregroundColor, value: UIColor.charcoalGrey, range: NSMakeRange(0, mutableString.length))
```
doctor3w: Oh I guess we do add attributes after. Nevermind the comment way above.

doctor3w (changes requested):
Looks good, just a few changes.

Podcast/EpisodeDetailViewController.swift
natashaarmbrust: I disagree. I feel like the header should not scroll and just the description should. Playing with the UI I think this looks much better. We can talk to eileen about it on Sunday and I will retroactively make changes if we want to change this.

Podcast/ExternalProfileViewController.swift
natashaarmbrust: We actually need this call and the result of it not calling it was the reason for the bug. Since this is the external profile that means the series isSubscribed is considered the isSubscribed of the external profile user and not the currentUser. Thus, we need to make this call to fetch the series again or else we won't have that isSubscribed currently reflecting the currentUser. One thing we could do is get it from the cache right. Well in the cache its stored with the isSubscribed state of the external profile user and not the current user. So when we request the series again it will update the cache for this series with the currentUsers state. We could also call a isUserSubscribed endpoint but unfortunately no such endpoint exists. Even if one did the only thing we would probably be saving here is a couple bits over the network.

PR #209, opened 2017-11-12
Lockscreen Player

doctor3w:
Pretty self-explanatory. No image artwork yet, something weird with Haneke need to discuss at meeting.


natashaarmbrust:
LGTM

PR #213, opened 2017-11-13
Lower priority for problematic height constraints

dantheli:
I strongly suspect constraint breaking is due to an inconsistency with UITableViews and UICollectionView row/item sizes and our programmatic constraints that create them.

A common practice is to lower the priority of created constraints to let iOS have the final say regarding the intrinsic size of our cells, often to 999. This way, iOS's encapsulated layout constraints with the default priority of 1000 have precedence, and no constraints end up broken at the end of the day.


mindylou (approved):
it works!! lgtm

natashaarmbrust:
I've seen this solution before as well. LGTM.

PR #227, opened 2017-11-16
Change image cache

dantheli:
Kingfisher > Haneke


Podcast/ImageView.swift
```diff
@@ -1,10 +1,11 @@
-
-import Haneke
+import Kingfisher
+import NVActivityIndicatorView
```
natashaarmbrust: why are you importing NVActivityIndicatorView

Podcast/ImageView.swift
```diff
@@ -1,10 +1,11 @@
...
 
 class ImageView: UIImageView {
     
     func setImageAsynchronouslyWithDefaultImage(url: URL?, defaultImage: UIImage = #imageLiteral(resourceName: "nullSeries")) {
         guard frame != .zero else { return }
```
natashaarmbrust: please remove frame != .zero

natashaarmbrust (changes requested):

PR #231, opened 2017-11-19
fixing listening history bug / pull to refresh bug

natashaarmbrust:
- Now listening history actually is paginated (bug before)
- Fixes all pull to refresh bugs that used to have a default boolean in the fetch endpoint request (apparently this default boolean will always be false when calling from a `#selector`)
- Refactors adding refresh control to inside `EmptyStateTableView`

### DIFF
--- a/Podcast/FeedViewController.swift
+++ b/Podcast/FeedViewController.swift
@@ -45,7 +45,7 @@ class FeedViewController: ViewController {
         feedTableView.rowHeight = UITableViewAutomaticDimension
         feedTableView.estimatedRowHeight = 200.0
         feedTableView.reloadData()
-        feedTableView.addInfiniteScroll { (tableView) -> Void in
+        feedTableView.addInfiniteScroll { _ -> Void in
             self.fetchFeedElements(isPullToRefresh: false)
         }
         //tells the infinite scroll when to stop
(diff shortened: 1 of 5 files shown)


Podcast/FeedViewController.swift
```diff
@@ -46,7 +46,7 @@ class FeedViewController: ViewController {
         feedTableView.estimatedRowHeight = 200.0
         feedTableView.reloadData()
         feedTableView.addInfiniteScroll { (tableView) -> Void in
```
mindylou: this can just be `{ tableView in`

Podcast/ListeningHistoryViewController.swift
```diff
@@ -115,20 +112,33 @@ class ListeningHistoryViewController: ViewController, UITableViewDelegate, UITab
     //MARK
     //MARK - Endpoint Requests
     //MARK
+    func emptyStateTableViewHandleRefresh() {
+        fetchEpisodes(refresh: true)
+    }
 
-    @objc func fetchEpisodes(refresh: Bool = true) {
+    func fetchEpisodes(refresh: Bool) {
```
mindylou: is `refresh` ever set to false? might be easier to have a default value of true if it's usually true

mindylou (commented):
looks good! just two comments/questions

PR #233, opened 2017-11-19
transfer search query to search itunes

mindylou:
- fixes ux of having to re-type search query
- automatically displays search controller w/keyboard active

### DIFF
--- a/Podcast/SearchITunesViewController.swift
+++ b/Podcast/SearchITunesViewController.swift
@@ -13,10 +13,16 @@ class SearchITunesViewController: ViewController, UITableViewDelegate, UITableVi
     
     var searchController: UISearchController!
     var tableView: EmptyStateTableView!
+    var initialQuery: String?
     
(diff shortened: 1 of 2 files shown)


Podcast/SearchITunesViewController.swift
```diff
@@ -13,10 +13,16 @@ class SearchITunesViewController: ViewController, UITableViewDelegate, UITableVi
     
     var searchController: UISearchController!
     var tableView: EmptyStateTableView!
+    var initialQuery: String?
     
     var searchResults: [Series] = []
     let seriesCellIdentifier = "SeriesCell"
     let seriesCellHeight: CGFloat = 95
+
+    convenience init(query: String?) {
```
natashaarmbrust: since this is a convenience init let's have query not be optional. If someone wanted to initailize this viewcontroller without a query they could just use the regular init

natashaarmbrust (approved):
LGTM just one line to change

PR #240, opened 2018-02-05
adding search Timer

natashaarmbrust:
- adds timer to delay searching

### DIFF
--- a/Podcast/SearchViewController.swift
+++ b/Podcast/SearchViewController.swift
@@ -95,7 +95,6 @@ class SearchViewController: ViewController, UISearchControllerDelegate, UITableV
     let searchITunesHeaderHeight: CGFloat = 79.5
     var tableViewData: [SearchDataSourceDelegate]!
     var hasLoaded: Bool = false
-    var searchResults: [Any] = []
     let tabBarUnderlineViewHeight: CGFloat = 44
     var didDismissItunesHeaderForQuery: Bool = false
     var lastSearchText: String = ""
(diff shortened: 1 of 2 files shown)


Podcast/SearchViewController.swift
```diff
@@ -240,21 +241,36 @@ class SearchViewController: ViewController, UISearchControllerDelegate, UITableV
...
-            data.fetchData(query: searchText, newSearch: true)
+        if self.lastSearchText == searchText && pastSearchesTableView.isHidden { return }
+        self.lastSearchText = searchText
+        self.tabUnderlineView.isHidden = false
+        self.searchResultsTableView.isHidden = false
+        self.pastSearchesTableView.isHidden = true
+        self.didDismissItunesHeaderForQuery = false
+        self.addSearchITunesHeader()
+        self.mainScrollView = self.searchResultsTableView
+        self.searchResultsTableView.startLoadingAnimation()
```
mindylou: why do these all use `self`?

Podcast/SearchViewController.swift
```diff
@@ -240,21 +241,36 @@ class SearchViewController: ViewController, UISearchControllerDelegate, UITableV
...
+
+        self.searchResultsTableView.reloadData()
+        self.searchResultsTableView.loadingAnimation.bringSubview(toFront: self.searchResultsTableView)
+        self.updateTableViewInsetsForAccessoryView()
+
+    }
+
+    @objc func searchAfterDelay() {
+        for data in self.tableViewData {
+            data.fetchData(query: lastSearchText, newSearch: true)
```
mindylou: the "no results found" screen shows up for a little bit after the loading indicator spins for a bit, and then shows results (only for series and episodes):
image

mindylou (commented):
looks good! just one small bug that I found with searching/null states

Podcast/SearchViewController.swift
natashaarmbrust: thanks for catching that! I had it in a completion block before

doctor3w (approved):
LGTM

PR #241, opened 2018-02-07
Added Fabric/Crashlytics

doctor3w:


Podcast/InternalProfileViewController.swift
```diff
@@ -7,6 +7,7 @@
 //
 
 import UIKit
+import Crashlytics
```
natashaarmbrust: is there a reason Crashlytics needs to be imported in this file? Doesn't look like we use it

Podfile.lock
```diff
@@ -1,5 +1,8 @@
 PODS:
   - Alamofire (4.5.1)
+  - Crashlytics (3.9.3):
+    - Fabric (~> 1.7.2)
```
natashaarmbrust, line 4: do we need this second Fabric (~> 1.7.2)? Does Crashlytics not work on Fabric 1.7.3?

natashaarmbrust (approved):
LGTM. two questions. Also is the username/password in "login pls"?

Podfile.lock
natashaarmbrust: nevermind

PR #254, opened 2018-02-22
Fixed tagString implementation

jackthmp:
Added case in which there is only 1 tag for a podcast

### DIFF
--- a/Podcast/Series.swift
+++ b/Podcast/Series.swift
@@ -32,7 +32,7 @@ class Series: NSObject {
             else {
                 for (i,tag) in tags.enumerated() {
                     if i == tags.count - 1 {
-                        tagString += ", and "
+                        tagString += tags.count == 2 ? " and " : ", and "
                     } else if i != 0 {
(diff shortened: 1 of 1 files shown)


Podcast/Series.swift
```diff
@@ -28,13 +28,16 @@ class Series: NSObject {
...
-            for (i,tag) in tags.enumerated() {
-                if i == tags.count - 1 {
-                    tagString += ", and "
-                } else if i != 0 {
-                    tagString += ", "
+            if tags.count == 1 { tagString += tags[0].name }
+            else {
+                for (i,tag) in tags.enumerated() {
+                    if i == tags.count - 1 {
+                        tagString += ", and "
```
natashaarmbrust: `tagString += tags.count == 2 ? " and " : ", and"` (I think that should work) so we take care of the case where we have two tags and don't want a comma

natashaarmbrust:
Make that one change then feel free to merge in

PR #255, opened 2018-02-23
Natasha/sharing

natashaarmbrust:
- Adds UI for sharing with specific followers
- We need to add backend functionality for this (backend calls are fake rn)


Podcast/SharedContentViewController.swift
```diff
@@ -0,0 +1,61 @@
...
+import NVActivityIndicatorView
+import SnapKit
+
+// View Controller showing all content personally shared with you by other users
+// TODO: refactor later on, quick hack for MVP
+class SharedContentViewController: FeedViewController {
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        title = "Shared With You"
```
natashaarmbrust: TODO: add different null state view

Podcast/FeedViewController.swift
```diff
@@ -28,14 +28,15 @@ class FeedViewController: ViewController {
     var feedMaxTime: Int = Int(Date().timeIntervalSince1970)
     var continueInfiniteScroll = true
     var feedSet: Set = Set<FeedElement>()
+    var emptyStateViewType: EmptyStateType = .feed
```
mindylou, line 31: shouldn't `FeedViewController`'s  empty state always be `.feed`?

Podcast/SeriesDetailHeaderView.swift
```diff
@@ -115,7 +115,7 @@ class SeriesDetailHeaderView: UIView, UICollectionViewDelegate, UICollectionView
         
         shareButton = UIButton(type: .custom)
         shareButton.adjustsImageWhenHighlighted = true
-        shareButton.setImage(#imageLiteral(resourceName: "shareButton"), for: .normal)
+        shareButton.setImage(#imageLiteral(resourceName: "iShare"), for: .normal)
```
mindylou, line 133: this asset isn't showing up when i pulled...?

Podcast/ShareEpisodeViewController.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        let tappedUser = users[indexPath.row]
+        let shareEpisodeOption = ActionSheetOption(type: .confirmShare, action: {
+            self.episode.share(with: tappedUser, success: { self.navigationController?.popViewController(animated: true)}, failure: nil) // TODO: error handling
+        })
+
+        var header: ActionSheetHeader?
+
+        let imageView = ImageView()
```
mindylou, line 38: does this still throw errors if initialized with no arguments?

Podcast/ShareEpisodeViewController.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        title = "Tap to Share"
+    }
+
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        let tappedUser = users[indexPath.row]
+        let shareEpisodeOption = ActionSheetOption(type: .confirmShare, action: {
+            self.episode.share(with: tappedUser, success: { self.navigationController?.popViewController(animated: true)}, failure: nil) // TODO: error handling
```
mindylou: it would be nice to have a brief notification/something pop up to show that it was a success/failure but i know that's a todo

Podcast/EpisodeDetailViewController.swift
```diff
@@ -75,13 +75,19 @@ class EpisodeDetailViewController: ViewController, EpisodeDetailHeaderViewDelega
     func episodeDetailHeaderDidPressMoreButton(view: EpisodeDetailHeaderView) {
         guard let episode = episode else { return }
         let option1 = ActionSheetOption(type: .download(selected: episode.isDownloaded), action: nil)
+        let shareEpisodeOption = ActionSheetOption(type: .shareEpisode, action: {
```
mindylou, line 78: this code looks like it's coming up a lot, you could refactor this into a specific `ActionSheetOption` but i know a lot of the action sheet options could be refactored so it might be easier to do them all at once

Podcast/InternalProfileViewController.swift
```diff
@@ -95,6 +95,8 @@ class InternalProfileViewController: ViewController, UITableViewDelegate, UITabl
         case (0,2): //subscriptions
             let subscriptionsViewController = SubscriptionsViewController()
             navigationController?.pushViewController(subscriptionsViewController, animated: true)
+        case (0,3): // shared content
+            navigationController?.pushViewController(SharedContentViewController(), animated: true)
```
mindylou: i feel like the shared content should go in an inbox from the feed, like how Instagram shows the DM inbox in the main feed instead of on profile?

Podcast/SharedContentViewController.swift
```diff
@@ -0,0 +1,70 @@
...
+//
+//  Created by Natasha Armbrust on 2/15/18.
+//  Copyright © 2018 Cornell App Development. All rights reserved.
+//
+
+import UIKit
+import NVActivityIndicatorView
+import SnapKit
+
+// View Controller showing all content personally shared with you by other users
```
mindylou: nice! use a triple slash (`///`) here so it comes up in documentation

mindylou (changes requested):
a few comments! otherwise this is good

Podcast/FeedViewController.swift
natashaarmbrust: Nope. For `SharingContentViewController` in profile I do a quick hack by subclassing `FeedViewController`.

Podcast/ShareEpisodeViewController.swift
natashaarmbrust: We actually never use the imageView only its image so its okay. Its a hack way to use KingFisher (which only works for ImageViews) to download the image async / use their image cache

Podcast/SharedContentViewController.swift
natashaarmbrust: done

Podcast/ShareEpisodeViewController.swift
natashaarmbrust: I can add this!

doctor3w:
Are we waiting to pull this until I do the backend?

natashaarmbrust:
Added facebook friends to settings instead of internal profile because it makes internal profile look much better. In addition, because we also have Facebook friend suggestions on feed I don't think its necessarily too important to have in internal profile as well.

jackthmp (approved):
Looks good to me!

PR #256, opened 2018-02-25
fix spacing and line height issues in episode descriptions

mindylou:
update spacing to conform to redlined designs: 

image

### DIFF
--- a/Podcast/EpisodeSubjectView.swift
+++ b/Podcast/EpisodeSubjectView.swift
@@ -176,16 +176,7 @@ class EpisodeSubjectView: UIView {
         // this is to avoid newlines/paragraphs showing up after truncating text
         let stringWithoutNewlines = episode.attributedDescription.string.replacingOccurrences(of: "\n", with: "")
         let mutableString = NSMutableAttributedString(string: stringWithoutNewlines)
-        let style = NSMutableParagraphStyle()
-        style.lineSpacing = 4
-        style.lineBreakMode = .byTruncatingTail
-        let attributes = [
-            NSAttributedStringKey.paragraphStyle: style,
-            NSAttributedStringKey.font: UIFont._14RegularFont(),
-            NSAttributedStringKey.foregroundColor: UIColor.charcoalGrey
-        ]
-        mutableString.addAttributes(attributes, range: NSMakeRange(0, mutableString.length))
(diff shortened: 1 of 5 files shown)


Podcast/EpisodeSubjectView.swift
```diff
@@ -173,7 +173,19 @@ class EpisodeSubjectView: UIView {
     func setupWithEpisode(episode: Episode) {
         episodeNameLabel.text = episode.title
         dateTimeLabel.text = episode.dateTimeLabelString
-        descriptionLabel.attributedText = episode.attributedDescription
+        // this is to avoid newlines/paragraphs showing up after truncating text
+        let stringWithoutNewlines = episode.attributedDescription.string.replacingOccurrences(of: "\n", with: "")
```
natashaarmbrust, line 177: is this the same code as above? If so, maybe include this as an extension to NSMutableString as a function that takes in an episode

Podcast/PlayerEpisodeDetailView.swift
```diff
@@ -102,11 +102,14 @@ class PlayerEpisodeDetailView: UIView, UIGestureRecognizerDelegate {
         episodeTitleLabel.text = episode.title
         dateLabel.text = episode.dateTimeLabelString
         let mutableString = NSMutableAttributedString(attributedString: episode.attributedDescription)
-        mutableString.addAttribute(NSAttributedStringKey.foregroundColor, value: UIColor.charcoalGrey, range: NSMakeRange(0, mutableString.length))
         let style = NSMutableParagraphStyle()
         style.lineSpacing = 4
-        mutableString.addAttribute(NSAttributedStringKey.paragraphStyle, value: style, range: NSMakeRange((0), mutableString.length))
-        mutableString.addAttribute(NSAttributedStringKey.font, value: UIFont._14RegularFont(), range: NSMakeRange(0, mutableString.length))
+        let attributes = [
```
natashaarmbrust: Seems like same for here so the comment above applies doubly

natashaarmbrust (changes requested):
LGTM .. maybe just one refactor

Podcast/EpisodeSubjectView.swift
mindylou: done! i made an extension to `NSAttributedString`

PR #262, opened 2018-02-27
fixed date parsing for last updated string

jackthmp:
I changed the date difference calc to be based off of the seconds since the date the podcast was updated to make the date last updated more accurate


natashaarmbrust:
LGTM!

PR #316, opened 2018-03-08
Jack/bugbash

jackthmp:
- Fixed app logo on login screen
- changed "Get Started" button color
- changed keyboard return button and added observer for textFieldDidChange on LoginUsernameViewController to check username availability as you type
- used safeAreaInsets to offset tab bar and action sheet
- changed null state internal profile subscriptions view


Podcast/ChangeUsernameView.swift
```diff
@@ -105,13 +110,19 @@ class ChangeUsernameView: UIView, UITextFieldDelegate {
...
+        usernameTextField.isEnabled = true
         successView.image = #imageLiteral(resourceName: "failure_icon")
         successView.isHidden = false
-        usernameTextField.isEnabled = true
+        continueButton.isEnabled = false
+        continueButton.backgroundColor = .silver
+    }
+    
+    @objc func textFieldDidChange(_ textField: UITextField) {
+        delegate?.changeUsernameViewTextFieldDidChange(changeUsernameView: self, username: textField.text!)
```
mindylou: does this throw errors if there's no text/someone deletes everything? or is it just the empty string

Podcast/InternalProfileViewController.swift
```diff
@@ -192,11 +193,21 @@ class InternalProfileViewController: ViewController, UITableViewDelegate, UITabl
             cell.selectionStyle = .gray
             return cell
         case subscriptionsTableView:
-            let cell = tableView.dequeueReusableCell(withIdentifier: reusableSubscriptionCellID, for: indexPath) as? SearchSeriesTableViewCell ?? SearchSeriesTableViewCell()
-            if let series = subscriptions?[indexPath.row] {
-                cell.configure(for: series, index: indexPath.row, showLastUpdatedText: true)
+            guard let subs = subscriptions else { return UITableViewCell() }
```
mindylou, line 196: should a `SearchSeriesTableViewCell` be returned here instead ?

Podcast/InternalProfileViewController.swift
```diff
@@ -233,9 +244,16 @@ class InternalProfileViewController: ViewController, UITableViewDelegate, UITabl
...
-                let seriesDetailViewController = SeriesDetailViewController(series: series)
-                navigationController?.pushViewController(seriesDetailViewController, animated: true)
+            guard let subs = subscriptions else { return }
+            if subs.count > 0 {
+                if let series = subscriptions?[indexPath.row] {
+                    let seriesDetailViewController = SeriesDetailViewController(series: series)
+                    navigationController?.pushViewController(seriesDetailViewController, animated: true)
+                }
+            }
+            else {
```
mindylou: put the `} else {` on one line

Podcast/ActionSheetViewController.swift
```diff
@@ -295,7 +298,7 @@ class ActionSheetViewController: UIViewController, UITableViewDataSource, UITabl
         UIView.animate(withDuration: animated ? 0.25 : 0.0) {
             
             self.darkBackgroundView.alpha = 0.8
-            self.actionSheetContainerView.frame = CGRect(x: 0, y: self.view.frame.height - self.actionSheetContainerView.frame.height, width: self.actionSheetContainerView.frame.width, height: self.actionSheetContainerView.frame.height)
+            self.actionSheetContainerView.frame = CGRect(x: 0, y: self.view.frame.height - self.actionSheetContainerView.frame.height - self.safeArea.bottom, width: self.actionSheetContainerView.frame.width, height: self.actionSheetContainerView.frame.height + self.safeArea.bottom)
```
mindylou, line 301: i know this is frame based but for future reference snapkit does have safe area insets built in!

https://github.com/SnapKit/SnapKit/issues/448

mindylou (approved):
LGTM

Podcast/InternalProfileViewController.swift
jackthmp: I don't want to return a SearchSeriesTableViewCell in the case that the subscriptions haven't been fetched yet (i.e. they are nil) because then there would just be a white cell

Podcast/ChangeUsernameView.swift
jackthmp: Hmm I'm not sure, I can try to change `username: textField.text!` to `username: textField.text ?? ""` to safely unwrap it?

jackthmp: We did force unwrap it in `textFieldDidEndEditing` so I think it might be okay.
@natashaarmbrust ??

mindylou: i think its fine then if it was used before!

natashaarmbrust: safely unwrap it and only call the delegate if it isn't nil and not the empty string

Podcast/InternalProfileViewController.swift
```diff
@@ -192,11 +193,21 @@ class InternalProfileViewController: ViewController, UITableViewDelegate, UITabl
...
+                if let series = subscriptions?[indexPath.row] {
+                    cell.configure(for: series, index: indexPath.row, showLastUpdatedText: true)
+                }
+                cell.backgroundColor = .offWhite
+                return cell
+            }
+            else {
+                guard let user = System.currentUser else { return UITableViewCell() }
+                let cell = NullProfileTableViewCell(user: user)
+                tableView.backgroundColor = .paleGrey
```
natashaarmbrust: why are we changing the background color here?

natashaarmbrust (approved):
LGTM

Podcast/InternalProfileViewController.swift
jackthmp: simulator screen shot - iphone x - 2018-03-08 at 23 09 52

the default color is white for the SearchSeriesTableViewCells

jackthomp: Oh if I just set it to paleGrey in viewDidLoad it doesn't show through the series cells

jackthmp: Oh I can just change the backgroundColor in viewDidLoad, I wasn't sure if the .offwhite that it was set to would show through

Issue #313, opened 2018-03-08
Add better null state for "Shared with you"


natashaarmbrust:
per @eileendai 's request

Issue #301, opened 2018-03-08
Remove filtering of current user logic from search


natashaarmbrust:
- this causes infinite scroll to stop 
- backend will filter on their end for search user

Issue #300, opened 2018-03-08
Fix app pushing backwards upon cancelling adding facebook account 


natashaarmbrust:
I hit “Cancel” in the UIAlert for adding a facebook account, the app pushes backwards, which isn’t my intended action

PR #317, opened 2018-03-09
Natasha/bugbash

### LINKED ISSUE #314: Change play pause button on player

natashaarmbrust:
fixes #314 , #313 , #309 , #303 , #301 , #300 , finally #315 , and finally finally #308


Podcast/UserSettings.swift
```diff
@@ -47,86 +47,134 @@ class UserSettings: NSObject {
...
                     // TODO: add error handling
+                    let presentAlert =  { self.present(UIAlertController.somethingWentWrongAlert(), animated: true, completion: nil) }
                     user.changeUsername(username: username, success: {
-                        let alert = UIAlertController(title: "Success", message: "Changed username to @\(username)", preferredStyle: .alert)
-                        alert.addAction(UIAlertAction(title: "Got it!", style: .default, handler: { _ in changeUsernameVC.navigationController?.popViewController(animated: true)}))
-                        changeUsernameVC.present(alert, animated: true, completion: nil)}, failure: {
-                        changeUsernameVC.present(UIAlertController.somethingWentWrongAlert(), animated: true, completion: nil)
-                    })
+                        self.present(UIAlertController.success(viewController: self, message: "Changed username to @\(username)"), animated: true, completion: nil)}, failure: presentAlert)
                 }, type: .textField("@username"))
```
mindylou: maybe do `"@\(username)"` to show what their current username is

Podcast/Authentication.swift
```diff
@@ -15,10 +15,48 @@ enum AuthenticationEndpointRequestType {
...
+    static func conversion(from error: Error?) -> SignInResult {
+        switch(error) {
+        case .none:
+            return SignInResult.success
+        case .some(let googleError):
+            switch(googleError.code) {
+            case GIDSignInErrorCode.canceled.rawValue, GIDSignInErrorCode.hasNoAuthInKeychain.rawValue:
+                return SignInResult.cancelled
+            default:
+                return SignInResult.failure
```
mindylou, line 45: not really sure what happened but i added my google account and it just took me to google lol and i couldn't really exit the view unless if i tapped on the bottom tab bar
img_664efb63a346-1

Podcast/Authentication.swift
```diff
@@ -41,31 +79,31 @@ class Authentication: NSObject, GIDSignInDelegate {
 
         GIDSignIn.sharedInstance().scopes.append(contentsOf: [profileScope, emailScope])
         facebookLoginManager = LoginManager()
+        facebookLoginManager.loginBehavior = .web
```
mindylou, line 82: i like this view much better than the old one!

mindylou: although kinda annoying to log in every time and not have Facebook remember

Podcast/Episode.swift
```diff
@@ -310,8 +310,17 @@ class Episode: NSObject, NSCoding {
     }
     
     func createRecommendation(success: ((Bool, Int) -> ())? = nil, failure: ((Bool, Int) -> ())? = nil) {
+        if let user = System.currentUser, let hasRecasted = user.hasRecasted, !hasRecasted { // first recast
```
mindylou, line 313: if i'm trying to have this pop up now do I have to make an account? because technically user defaults haven't been written in yet but i've recasted stuff

mindylou (commented):
i'm getting the one weird bug with google but otherwise looks good!

Podcast/Episode.swift
natashaarmbrust: It's specific to the phone and resets on logout .. so every time you log in you would get shown it .. I know @drewsdunne  mentioned that the whole logout reset might get annoying

Podcast/Authentication.swift
natashaarmbrust: Yea the app behavior (over login) is buggier which is why I chose this one. The annoying this is it not remembering your password. Let me play again with the other one but I was having issues when it wasn't web

Podcast/Authentication.swift
natashaarmbrust: Yes I know this is the only current bug :( ... I commented where it is because I think its a Google SDK issue and not our issue.

natashaarmbrust: Yea both are very buggy (when I change Facebook from the web sign in behavior)  .. maybe I can ask @dantheli but im struggling with login here. It's an issue with completion blocks not being called and views (such as the one above) not popping off like they should

Podcast/Authentication.swift
mindylou: ok, I don't imagine many people signing out of the app too often so it shouldn't be a super big issue

Podcast/Episode.swift
doctor3w: Yeah I would just do once per install. Like once in UserDefaults it's there forever. I feel like I never log out of my apps and if I do it's by accident, and would not want to go through on boarding again, which includes this recasting demo.

doctor3w (commented):
LGTM

Issue #332, opened 2018-03-13
Look into listening history saving 


natashaarmbrust:
I've noticed my listening duration on an episode lags a bit from when I remember I was when opening the app back up again and starting up a podcast I was previously listening to. Anyone else experience this issue?

mindylou:
i've left the app open on my phone overnight and then it doesn't save the listening history at all, not sure if that's the same issue though

PR #345, opened 2018-03-16
Natasha/updates

### LINKED ISSUE #341: Change new users to discover at first

natashaarmbrust:
- fixes #341, #332, #325, #324, #321
- refactors action sheet
- also adds about page in settings so we can advertise ourselves!


jackthmp:
An episode wasn't playing so I tried to play another episode and the app crashed and I got this:
`*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'Invalid number value (NaN) in JSON write'`
I'm guessing this probably happened when you tried to save the current progress of the episode and the progress was nil? I'm not sure

Podcast/AppDelegate.swift
```diff
@@ -67,10 +67,18 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         tabBarController.accessoryViewController?.collapseAccessoryViewController(animated: animated)
         tabBarController.showTabBar(animated: animated)
     }
+
+    // upon episode play, show mini player and animate to player
+    func showAndExpandPlayer() {
+        showPlayer(animated: false)
+        DispatchTime.waitFor(milliseconds: 500) {
```
jackthmp: Why do we wait for 500 milliseconds? I feel like the player should seem to immediately open instead of having that brief moment where you see the mini player

Podcast/Assets.xcassets/logo.imageset/Contents.json
```diff
@@ -1,23 +1,23 @@
 {
-  "info" : {
-    "author" : "zeplin",
-    "version" : 1
-  },
   "images" : [
     {
       "idiom" : "universal",
-      "filename" : "group4.png",
+      "filename" : "bitmap.jpg",
```
jackthmp, line 5: Are we updating the app icon to this too?

jackthmp (changes requested):
LGTM except for a few comments and a crash :(

Podcast/Assets.xcassets/logo.imageset/Contents.json
natashaarmbrust: yes im going to add the app icons to this PR!

Podcast/AppDelegate.swift
natashaarmbrust: you are right that looks much better!

natashaarmbrust:
@jackthmp I can't reproduce the Exception error but I did change the listening duration saving to check for nans so we shouldn't get that exception anymore! If you have any more info on the bug or know how to reproduce lmk

Podcast/Player.swift
```diff
@@ -174,7 +184,7 @@ class Player: NSObject {
             if currentItem.status == .readyToPlay {
                 savedRate = rate
                 player.pause()
-                updateCurrentPercentageListened()
+                //updateCurrentPercentageListened()
```
doctor3w, line 190: Need this still?

Podcast/Player.swift
```diff
@@ -133,13 +137,19 @@ class Player: NSObject {
         // cleanup any previous AVPlayerItem
         pause()
         removeCurrentItemStatusObserver()
-        setCurrentEpisodeToPreviouslyPlayingEpisode()
 
+        if let listeningDuration = listeningDurations[episode.id] { // if we've played this episode before
+            currentTimeAt = listeningDuration.currentProgress
+        } else {
+            listeningDurations[episode.id] = ListeningDuration(id: episode.id, currentProgress: episode.currentProgress, percentageListened: 0, realDuration: nil)
+            currentTimeAt = episode.currentProgress
```
doctor3w, line 145: Do we have two current progress's to update now?

doctor3w (commented):
LGTM

Podcast/Player.swift
natashaarmbrust: nope

PR #357, opened 2018-03-21
fixed user followers number update bug

doctor3w:
Just small changes I pulled out of other work I'm doing for downloads.

### DIFF
--- a/Podcast/UserDetailHeaderView.swift
+++ b/Podcast/UserDetailHeaderView.swift
@@ -64,12 +64,13 @@ class UserDetailHeaderView: UIView {
             contentContainerTop = make.top.equalToSuperview().constraint
             make.leading.equalToSuperview()
             make.trailing.equalToSuperview()
-            make.bottom.equalToSuperview()
+            make.bottom.equalToSuperview().priority(.required)
         }
         
         infoAreaView.snp.makeConstraints { make in
             make.height.greaterThanOrEqualTo(UserDetailHeaderView.infoAreaMinHeight)
-            make.top.greaterThanOrEqualTo(containerView.snp.top)
+            make.top.equalToSuperview()
+//            make.top.greaterThanOrEqualTo(containerView.snp.top)
             make.leading.trailing.equalToSuperview()
             make.bottom.equalTo(buttonBar.snp.top)
         }
@@ -89,7 +90,7 @@ class UserDetailHeaderView: UIView {
         subscriptionsView.snp.makeConstraints { make in
             make.height.equalTo(UserDetailHeaderView.subscriptionsViewHeight)
             make.leading.trailing.bottom.equalToSuperview()
-            make.bottom.equalToSuperview()
+            make.bottom.equalToSuperview().priority(.required)
         }
         
     }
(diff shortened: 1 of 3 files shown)


Podcast/UserDetailViewController.swift
```diff
@@ -83,20 +84,21 @@ final class UserDetailViewController: ViewController {
...
-            make.height.equalTo(UIApplication.shared.statusBarFrame.height + (navigationController?.navigationBar.frame.height)!)
+            make.height.equalTo(UIApplication.shared.statusBarFrame.height + navBarHeight)
         }
         
     }
     
     func scrollViewDidScroll(_ scrollView: UIScrollView) {
         // updates the constraint to stretch the header view's green background
         let offset = max(0, -(scrollView.contentOffset.y + scrollView.adjustedContentInset.top))
+        print(offset)
```
natashaarmbrust: remove print

Podcast/UserDetailViewController.swift
```diff
@@ -63,7 +64,7 @@ final class UserDetailViewController: ViewController {
         userDetailHeaderView = UserDetailHeaderView(frame: CGRect(x: 0, y: 0, width: view.frame.width, height: UserDetailHeaderView.minHeight))
         profileTableView.tableHeaderView = userDetailHeaderView
         userDetailHeaderView.snp.makeConstraints { make in
-            make.edges.equalToSuperview()
+//            make.edges.equalToSuperview()
```
natashaarmbrust, line 67: remove comment

Podcast/UserDetailHeaderView.swift
```diff
@@ -64,12 +64,13 @@ class UserDetailHeaderView: UIView {
...
             make.trailing.equalToSuperview()
-            make.bottom.equalToSuperview()
+            make.bottom.equalToSuperview().priority(.required)
         }
         
         infoAreaView.snp.makeConstraints { make in
             make.height.greaterThanOrEqualTo(UserDetailHeaderView.infoAreaMinHeight)
-            make.top.greaterThanOrEqualTo(containerView.snp.top)
+            make.top.equalToSuperview()
+//            make.top.greaterThanOrEqualTo(containerView.snp.top)
```
natashaarmbrust: remove comment

Podcast/UserDetailViewController.swift
```diff
@@ -19,6 +19,7 @@ final class UserDetailViewController: ViewController {
     
     let recastsHeaderViewTag = 2112
     let subscriptionsHeaderViewTag = 3110
+    let navBarHeight: CGFloat = 44
```
natashaarmbrust, line 22: let's make navBarHeight have a getter 
`get {
return navigationController?.navigationBar.frame.height ?? navBarHeight
}` so that we don't need to do the ?? everywhere but also have the correct height

natashaarmbrust (changes requested):

Podcast/UserDetailViewController.swift
doctor3w: I'm actually going to remove that instance of navigationController?.navigationBar.frame.height because everything else is calculated to 44 anyway, so if they ever changed the height (like with iOS 11) it's going to be weird.

PR #363, opened 2018-03-22
Natasha/nav bar

natashaarmbrust:
- adds discover headers to top of nav bar 
screen shot 2018-03-22 at 12 40 12 am

### DIFF
--- a/Podcast/DiscoverTopicViewController.swift
+++ b/Podcast/DiscoverTopicViewController.swift
@@ -12,20 +12,18 @@ import SnapKit
 /// Displays the discoverable content (series, episodes) for a given topic. 
 class DiscoverTopicViewController: DiscoverComponentViewController {
 
-    var topicImageView: UIImageView!
-    var topicLabel: UILabel!
     var relatedTopicsView: TrendingTopicsView!
     var episodesHeaderView: DiscoverCollectionViewHeaderView!
     var topEpisodesTableView: UITableView!
     var seriesHeaderView: DiscoverCollectionViewHeaderView!
     var topSeriesCollectionView: UICollectionView!
+    var topicLabel: UILabel!
 
     let episodesReuseIdentifier = "episodes"
     let seriesReuseIdentifier = "series"
     let episodesHeaderTag = 1
     let seriesHeaderTag = 2
     let collectionViewHeight: CGFloat = 160
-    let imageViewHeight: CGFloat = 84
     let relatedTopicsHeight: CGFloat = 150
 
     var relatedTopics = [Topic]()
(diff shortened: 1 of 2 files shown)


natashaarmbrust:
sorry I made this off of Jack's branch accidentally .. ignore all changes not in `DiscoverTopicViewController` and `ViewController` 
aka just this commit https://github.com/cuappdev/podcast-ios/pull/363/commits/2f475bcaf754b13623381cf5ba5f806a38dd5e62

Podcast/DiscoverTopicViewController.swift
```diff
@@ -115,17 +101,17 @@ class DiscoverTopicViewController: DiscoverComponentViewController {
...
         }
 
         topEpisodesTableView.tableHeaderView = headerView
         // adjust header height
-        let headerViewHeight = imageViewHeight + collectionViewHeight + 2 * headerHeight + relatedTopicsHeight
+        let headerViewHeight = collectionViewHeight + 2 * headerHeight + relatedTopicsHeight
         headerView.snp.makeConstraints { make in
             make.width.top.centerX.equalToSuperview()
-            make.height.equalTo(headerViewHeight)
+            //make.height.equalTo(headerViewHeight)
```
jackthmp: delete comment and declaration of headerViewHeight (never used)

Podcast/DiscoverTopicViewController.swift
```diff
@@ -170,10 +142,39 @@ class DiscoverTopicViewController: DiscoverComponentViewController {
...
+        navigationController?.navigationBar.setBackgroundImage(image.resizableImage(withCapInsets: UIEdgeInsets.zero, resizingMode: .stretch), for: .default)
+
+        navigationController?.navigationBar.tintColor = .offWhite
+    }
+
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        navigationItem.titleView = nil
+        navigationController?.navigationBar.setBackgroundImage(nil, for: .default)
+    }
```
jackthmp, line 165: So this sometimes lags and doesn't change the background image right away. Use `override func willMove(toParentViewController parent: UIViewController?) ` to immediately clear it when the back button is pressed.

jackthmp (approved):
LGTM! just a few changes

Podcast/DiscoverTopicViewController.swift
jackthmp: Hmm well actually if you use that then when you press the back button in a subtopic view controller it will clear the nav bar for a sec and then re-insert the image... so maybe check if the parentTopic is nil

natashaarmbrust: this doesn't work so I am going to keep it this way for now

PR #366, opened 2018-03-23
make lastUpdated an optional

mindylou:
- if there's no episodes, it's never been updated
- replacement series description text is "Never updated"


jackthmp (approved):
LGTM

PR #367, opened 2018-03-23
fix small status bar action sheet color bug

mindylou:
- when the action sheet was presented the status bar didn't properly dim
img_c2fa4d851ca4-1


jackthmp (changes requested):
The status bar and darkBackgroundView don't fade in sync... is there a way to extend the dark background view to go over the status bar?

jackthmp (approved):

jackthmp:
Looks good now!

Issue #365, opened 2018-03-22
DO NOT LEARN SPANISH


natashaarmbrust:
Daniel Li do your magic

dantheli:
See PR #379

PR #379, opened 2018-03-27
Remove fabric keys, google info plist, and facebook app id from repository

dantheli:
This PR removes private team related data (but not the git history). Google sign in's plist has been untracked, and Fabric keys and facebook app id are moved into Keys.plist:
```
facebook-app-id : string
fabric-api-key : string
fabric-build-secret : string
```

Address #365


natashaarmbrust (approved):
LGTM

PR #390, opened 2018-03-28
adding corner radius iOS 11

natashaarmbrust:
- adds a corner radius to artwork images


jackthmp (changes requested):
Need to add radius in EpisodeDetailHeaderView

PR #393, opened 2018-04-03
Mindy/tabbar

mindylou:
- redo tab bar to subclass `UITabBarController`
- tab tab bar to scroll up / pop to root view controller
- tap to scroll up is a bit buggy due to auto resizing table view cells/content offset


Podfile.lock
```diff
@@ -1,108 +0,0 @@
-PODS:
```
doctor3w: AHH don't delete this file!

doctor3w (approved):
LGTM

doctor3w:
Just get rid of that massive separator LOL

PR #399, opened 2018-04-11
Mindy/discover episodes bug

### LINKED ISSUE #398: Discover showing episodes twice
These episodes reappear when scrolling in discover 

img_fa4e91a6b8f0-2
img_fa4e91a6b8f0-1

mindylou:
- fixes #398 by only reloading data on view will appear (instead of previously re-fetching everything)
- adds pull to refresh in discover

### DIFF
--- a/Podcast/DiscoverComponentViewController.swift
+++ b/Podcast/DiscoverComponentViewController.swift
@@ -10,10 +10,6 @@ import UIKit
 import SnapKit
 import NVActivityIndicatorView
 
-protocol DiscoverTableViewDelegate: class {
-    func handlePullToRefresh()
-}
-
 /// ViewController helper subclass that creates the main UI components of the Discover user and topic views.
 class DiscoverComponentViewController: ViewController, NVActivityIndicatorViewable {
 
@@ -27,8 +23,6 @@ class DiscoverComponentViewController: ViewController, NVActivityIndicatorViewab
     var pageSize: Int { get { return 40 }} 
     var offset = 0
     var continueInfiniteScroll = true
-
-    weak var tableViewDelegate: DiscoverTableViewDelegate?
 
     override func viewDidLoad() {
         super.viewDidLoad()
@@ -74,7 +68,12 @@ class DiscoverComponentViewController: ViewController, NVActivityIndicatorViewab
     }
 
     @objc func pullToRefresh() {
-        tableViewDelegate?.handlePullToRefresh()
+        handlePullToRefresh()
+    }
+
+    /// Override this method when subclassing to implement a pull to refresh for the tableview.
+    func handlePullToRefresh() {
+
     }
 
 }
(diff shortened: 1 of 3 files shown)


Podcast/DiscoverComponentViewController.swift
```diff
@@ -62,4 +67,13 @@ class DiscoverComponentViewController: ViewController, NVActivityIndicatorViewab
         return header
     }
 
+    @objc func pullToRefresh() {
```
doctor3w: Why can't we just override this method?

Podcast/DiscoverTopicViewController.swift
```diff
@@ -175,16 +177,21 @@ class DiscoverTopicViewController: DiscoverComponentViewController {
         navigationController?.navigationBar.setBackgroundImage(nil, for: .default)
     }
 
-    func configureTopic() {
+    func configureTopic(isPullToRefresh: Bool = false) {
```
doctor3w: Better name: canPullToRefresh. Also any reason we need the word topic in this function name? What about just "setup" or "configure".

Podcast/DiscoverViewController.swift
```diff
@@ -121,11 +121,15 @@ class DiscoverViewController: DiscoverComponentViewController {
...
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
-        fetchDiscoverElements()
+        topSeriesCollectionView.reloadData()
+        topEpisodesTableView.reloadData()
     }
 
-    func fetchDiscoverElements() {
-        topSeriesCollectionView.reloadData()
+    func fetchDiscoverElements(isPullToRefresh: Bool = false) {
```
doctor3w: Same here.

doctor3w (changes requested):

Podcast/DiscoverComponentViewController.swift
mindylou: i had that before and it was throwing an error before but now it's not... weird

Podcast/DiscoverTopicViewController.swift
```diff
@@ -220,18 +226,20 @@ class DiscoverTopicViewController: DiscoverComponentViewController {
         relatedTopicsView.collectionView.reloadData()
     }
 
-    func fetchEpisodes(id: Int) {
+    func fetchEpisodes(id: Int, canPullToRefresh: Bool = false) {
 
         let topEpisodesForTopicEndpointRequest = DiscoverTopicEndpointRequest(requestType: .episodes, topicID: id, offset: offset, max: pageSize)
```
natashaarmbrust: let's reset the offset here instead of setup because this endpoint is what uses the offset

natashaarmbrust (commented):
LGTM

PR #406, opened 2018-05-04
Natasha/quick fixes

natashaarmbrust:
- Adds large titles

PLEASE TEST.


Podcast/EmptyStateView.swift
```diff
@@ -53,7 +53,7 @@ enum EmptyStateType {
         case .pastSearch:
             return "Find your favorite podcast episodes, series, & friends."
         case .bookmarks:
-            return "You can save podcast episodes for later here. Start looking now!"
+            return "" //You can save podcast episodes for later here. Start looking now!"
```
mindylou: why is there no empty state text for bookmarks?

Podcast/ViewController.swift
```diff
@@ -20,24 +20,32 @@
...
         if System.isiPhoneX() { insetPadding = iPhoneXBottomOffset }
         
         navigationItem.backBarButtonItem = UIBarButtonItem(title: "", style: .plain, target: nil, action: nil)
+        stylizeNavBar()
     }
     
     func stylizeNavBar() {
         navigationController?.navigationBar.tintColor = .sea
-        navigationController?.navigationBar.setBackgroundImage(nil, for: .default)
+        //navigationController?.navigationBar.setBackgroundImage(UIColor.offWhite.as1ptImage(), for: .default)
```
mindylou: remove comment

Podcast/ViewController.swift
```diff
@@ -65,18 +73,32 @@ class ViewController: UIViewController {
...
+
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+         displayNavTitle()
     }
 
     override func didMove(toParentViewController parent: UIViewController?) {
         super.didMove(toParentViewController: parent)
-        stylizeNavBar()
+        //stylizeNavBar()
```
mindylou: remove comment

mindylou (commented):
looks mostly good! i'll post a couple of screenshots of bugs i've found

mindylou:
the status bar color is white vs the nav bar is off white

img_1440

there's no bottom grey line under this nav bar anymore
img_1443

this transition scrolling down doesn't match with the underline bar for all of the tabs... here's an example
img_1442

natashaarmbrust:
The issue with the nav bar and status bar differing colors is with making the nav bar translucent .. its not actually a different color its just translucent so you see the background of it. I tried changing this and it cause a lot of bugs and renders the refresh control unseeable. I don't think its that huge of an issue for the resulting bugs when fixing it

natashaarmbrust:
The shadow image is another [iOS 11 bug](https://forums.developer.apple.com/thread/86828) 

The shadow image does not adjust properly when going from not using large titles to using large titles. Our problem is we switch between using large titles / not on the same navigation controller 

screen shot 2018-05-04 at 6 44 58 pm

Podcast/EmptyStateView.swift
natashaarmbrust: It looked weird with the null state and jump back in .. I'm adding in the empty state back in but removing the image so it looks better.

jackthmp (approved):
Is it possible to make the statusbar translucent? if you look really close the nav bar is a little translucent but then the status bar is just white.
could we possibly make the nav bar a little more translucent to match the opacity of the tab bar?

natashaarmbrust:
I've tried that @jackthmp . The problem is when I do then sometimes we get a bug where it now is grey because its taking on the view controllers background image.

Which is the backgroundColor not extending up enough. I can fix that with a background image but then the refresh control isn't visible / the large titles now take up more room below the background image. This shit is so buggy 👎

Podcast/EmptyStateView.swift
```diff
@@ -79,8 +79,8 @@ enum EmptyStateType {
         switch self {
         case .pastSearch:
             return #imageLiteral(resourceName: "searchIcon")
-        case .bookmarks:
-            return #imageLiteral(resourceName: "bookmark_empty_state")
+//        case .bookmarks:
+//            return #imageLiteral(resourceName: "bookmark_empty_state")
```
doctor3w, line 86: Why commented?

Podcast/InternalProfileViewController.swift
```diff
@@ -133,10 +131,6 @@ class InternalProfileViewController: ViewController, UITableViewDelegate, UITabl
     
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
-    }
-    
-    override func viewDidAppear(_ animated: Bool) {
-        super.viewDidAppear(animated)
         fetchSubscriptions()
```
doctor3w, line 134: I think it'd be better to fetch stuff and do API requests in viewDidAppear, WillAppear is usually for view hiding, moving, etc. I've read this somewhere else but I can't find it but here's a SO post that reiterates https://stackoverflow.com/questions/5562938/looking-to-understand-the-ios-uiviewcontroller-lifecycle

doctor3w (changes requested):
Looks good just two q's

Podcast/InternalProfileViewController.swift
natashaarmbrust: got it will change to be in viewDidAppear

Podcast/EmptyStateView.swift
natashaarmbrust: in case we want to use this image again, I don't want us to lose the image. It looks bad though with jump back in and the image so that's why its no longer showed.

PR #421, opened 2018-08-30
Added episode protocols

doctor3w:
Added a display view protocol to unify our cell setup for episode cells. Ignore the data source for now, it's not used just yet but more as an example for now. 

This is implemented on the BookmarkViewController.


jackthmp (dismissed):
LGTM, are you planning on deleting `func setup` in `BookmarkTableViewCell`?

doctor3w:
Only if BookmarkViewController is the only vc that uses it. But I have a feeling it isn't. Mindy and I's next job anyway is going through and putting this as many places as we can anyway and remove setup functions so I can just do it then.

Podcast/BookmarkViewController.swift
```diff
@@ -159,10 +159,33 @@ extension BookmarkViewController: UITableViewDataSource {
...
+        cell.displayView.set(seriesTitle: episode.seriesTitle)
+        cell.displayView.set(title: episode.title)
+        if let url = episode.smallArtworkImageURL {
+            cell.displayView.set(smallImageUrl: url)
+        }
+        if let url = episode.largeArtworkImageURL {
+            cell.displayView.set(largeImageUrl: url)
+        }
+        
+        cell.displayView.set(topics: episode.topics.map { topic in topic.name })
```
AAAstorga, line 178: nice use of map

Podcast/EpisodeProtocols.swift
```diff
@@ -0,0 +1,92 @@
...
+    func set(numberOfRecasts: Int)
+    
+    func set(downloadStatus: DownloadStatus)
+}
+
+class EpisodeTableDataSource: NSObject {
+    
+    // Can later adapt this to multiple cell types (like series and episodes)
+    private var sections: Int = 1
+    var identifier: String = "my_cell"
```
AAAstorga: can this be a better identifier

Podcast/EpisodeProtocols.swift
```diff
@@ -0,0 +1,92 @@
...
+    func set(recastBlurb: String)
+    func set(numberOfRecasts: Int)
+    
+    func set(downloadStatus: DownloadStatus)
+}
+
+class EpisodeTableDataSource: NSObject {
+    
+    // Can later adapt this to multiple cell types (like series and episodes)
+    private var sections: Int = 1
```
AAAstorga: don't need type definition here

Podcast/EpisodeProtocols.swift
AAAstorga: also use type inference

AAAstorga (dismissed):
I approve, but some minor style changes! Looks great. Love all the work being put in for this refactor!

Podcast/EpisodeProtocols.swift
doctor3w: oh this isn't used for anything rn, I was just playing around to see how I could abstract this more. I just wanted a test identifier to write out the code. Same with section

PR #424, opened 2018-09-01
Jack/endpoints

jackthmp:
Combined similarly functioning `EndpointRequest`s into one request.

### DIFF
--- a/Podcast/FollowsEndpointRequests.swift
+++ b/Podcast/FollowsEndpointRequests.swift
@@ -8,6 +8,7 @@
 
 import Foundation
 import SwiftyJSON
+import Alamofire
 
 enum UserFollowsType {
     case Followers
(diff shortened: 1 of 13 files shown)


Podcast/FollowsEndpointRequests.swift
```diff
@@ -54,31 +55,15 @@ class FetchUserFollowsByIDRequest: EndpointRequest {
...
-    }
-}
-
-class UnfollowUserEndpointRequest: EndpointRequest {
-    
-    // ID to unfollow
-    var userID: String
-    
-    init(userID: String) {
+    init(userID: String, method: HTTPMethod) {
```
mindylou: not sure if we should be passing the HTTP method here directly because it's not 100% understandable from a reader's perspective (ie. "post" a follow doesn't really make sense), maybe an enum called `FollowAction` or something like that could be useful here?

mindylou (changes requested):
Overall looks good! But maybe we should consider using enums as the methods to the endpoint requests instead of the HTTP methods themselves

Podcast/FollowsEndpointRequests.swift
AAAstorga: +1

Podcast/ShareEndpointRequests.swift
```diff
@@ -49,35 +50,26 @@ class FetchSharesEndpointRequest: EndpointRequest {
...
+        httpMethod = method
+        path = "/shares/\(id)/"
         
-        httpMethod = .delete
-        path = "/shares/\(shareId)/"
+        if method == .post {
+            self.userSharedWithIds = userSharedWithIds
+            
+            var shareString = ""
+            for (i,id) in userSharedWithIds.enumerated() {
```
AAAstorga: I think it would be cleaner here to do `userSharedWithIds.enumerated().forEach {...` but it is a personal preference

AAAstorga (dismissed):
I agree with Mindy, not sure we should be passing in HTTPMethod

Podcast/FollowsEndpointRequests.swift
jackthmp: I agree, I originally had an enum `Action` that had `create` or `delete` and an `httpMethod` var. I'll change it back!

doctor3w (approved):
LGTM. I do eventually want to change up the networking system, but for now this will reduce a lot of stupidly long named endpoints that are different by only the httpMethod.

mindylou (approved):

PR #428, opened 2018-09-11
Mindy/seriesdisplayview

### LINKED ISSUE #427: Refactor series cells to use new protocols
**Description:**
Refactor all cells displaying series to use `SeriesDisplayView` to remove duplicate cell configuration code.

mindylou:
- Closes #427 
- Looking for specific feedback for protocol methods (eg. `hideSubscribedButton`, grouping setting number of subscribers and isSubscribe) because some view controllers use the same cells but display different components or configure buttons using combined helper methods

### DIFF
--- a/Podcast/SeriesGridCollectionViewCell.swift
+++ b/Podcast/SeriesGridCollectionViewCell.swift
@@ -8,7 +8,7 @@
 import SnapKit
 import UIKit
 
-class SeriesGridCollectionViewCell: UICollectionViewCell {
+class SeriesGridCollectionViewCell: SeriesDisplayCollectionViewCell {
 
     let headerOffset: CGFloat = 60
     let imageTitlePadding: CGFloat = 8
@@ -20,6 +20,7 @@ class SeriesGridCollectionViewCell: UICollectionViewCell {
 
     override init(frame: CGRect) {
         super.init(frame: frame)
+        displayView = self
 
         isUserInteractionEnabled = true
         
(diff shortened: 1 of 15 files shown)


Podcast/SeriesGridCollectionViewCell.swift
```diff
@@ -63,16 +64,27 @@ class SeriesGridCollectionViewCell: UICollectionViewCell {
...
+    }
+
+    func set(largeImageUrl: URL) {
+        imageView.setImageAsynchronouslyWithDefaultImage(url: largeImageUrl)
+    }
+
+    func set(smallImageUrl: URL) {}
+    func hideSubscribedButton() {}
+    func set(author: String) {}
+    func set(topicsString: String) {}
```
myo3, line 89: I think the code is well-organized and I like how you guys are trying to centralize you cell set-up code. One thing that I don't like is how there's a lot of empty, unimplemented methods with this protocol. I feel like the protcol `SeriesDisplayView` in this example (`SeriesGridCollectionViewCell`) should describe exactly what methods one needs to "set" to set up the `SeriesGridCollectionViewCell`. B/c otherwise someone can just call `set(smallImageUrl: URL)` on a `SeriesGridCollectionViewCell` but that does nothing to a `SeriesGridCollectionViewCell` and therefore the code shouldn't allow `set(smallImageUrl: URL)` to be called since that's misleading. 

As to a solution, I'm still wrangling w it myself. I guess you could make the methods that not all cells implement optional. Alternatively you could have a base protocol that has methods that all cells implement and have more specific protocols inherit from that base protocol. And maybe use the visitor pattern to distinguish between different protocol types. At the end of the day, I feel the code should write in such a way that it is unambiguous on exactly what setter methods need to be called in `cellForRowAt` and the codes shouldn't allow the calling of setter methods that have empty implementations from every cell's `displayView`

PR #436, opened 2018-09-16
Added a basic search endpoint, started parsing.

doctor3w:
Added the Draft pod for endpoints, started parsing code by adding FeedKit. I think our best move is to make our own parsing, this way we can parse directly into our own models.


Recast/Models.swift
```diff
@@ -0,0 +1,239 @@
...
+    
+    private func parse(items: [RSSFeedItem]?) -> [Episode] {
+        guard let items = items else { return [] }
+        // TODO: Implement this
+        return []
+    }
+    
+    func parse(shouldParseItems: Bool) {
+        let parser = FeedParser(URL: partial.feedUrl)
+        parser.parseAsync(queue: DispatchQueue.global(qos: .userInitiated)) { (result) in
```
mindylou: remove parentheses

Recast/SearchExampleViewController.swift
```diff
@@ -0,0 +1,37 @@
...
+        SearchEndpoint(parameters: ["term": "#DORK", "media": "podcast", "limit": 1]).run()
+            .success { (response: SearchResults) in
+                guard response.resultCount > 0,
+                    let partial = response.results.first else { return }
+                Podcast.loadFull(from: partial, success: { podcast in
+                    print(podcast.title!)
+                    print(podcast.description!)
+                    print(podcast.link!)
+                    print(podcast.collectionName ?? "")
+                    print(podcast.items[1].title ?? "")
```
mindylou: is there a safer way to get stuff out of `items` aside from indexing into the array? is it guaranteed to have the 1st index?

mindylou: also, remove these print statements when you merge in

Recast/iTunesPodcasts/Models/RSS/mapAttributes.swift
```diff
@@ -0,0 +1,197 @@
...
+    ///   - path: The path of feed's element.
+    func map(_ attributes: [String : String], for path: RSSPath) {
+
+        switch path {
+            
+        case .rssChannelItem:
+
+            self.items.append(Episode())
+
+//        case .rssChannelImage:
```
mindylou: remove comments if these fields are unnecessary

Recast/iTunesPodcasts/Models/RSS/mapCharacters.swift
```diff
@@ -0,0 +1,114 @@
...
+        case .rssChannelCategory:                                   self.categories?
+            .append(string)
+        case .rssChannelGenerator:                                  self.generator                                                  = self.generator?.appending(string) ?? string
+        case .rssChannelDocs:                                       self.docs                                                       = self.docs?.appending(string) ?? string
+        case .rssChannelRating:                                     self.rating                                                     = self.rating?.appending(string) ?? string
+        case .rssChannelTTL:                                        self.ttl                                                        = Int(string)
+        case .rssChannelImageURL:                                   self.image                                                 = URL(string: string)
+//        case .rssChannelImageURL:                                   self.image?.url                                                 = self.image?.url?.appending(string) ?? string
```
mindylou: just a question... what's the difference between `image` and `image.url` in this case?

Recast/iTunesPodcasts/Models/iTunes/iTunesNamespace.swift
```diff
@@ -0,0 +1,276 @@
...
+    /// Use the <itunes:episode> tag in conjunction with the <itunes:season> tag
+    /// to indicate the order an episode should be presented within a season.
+    public var episode: Int?
+}
+
+// MARK: - Equatable
+//
+//extension ITunesNamespace: Equatable {
+//
+//    public static func ==(lhs: ITunesNamespace, rhs: ITunesNamespace) -> Bool {
```
mindylou: delete this if we don't need it

mindylou (changes requested):
looks good! just remove some of the commented out code if we don't need it to reduce clutter then i'll approve

Recast/iTunesPodcasts/Models/RSS/mapCharacters.swift
doctor3w: Image in RSS can contain more subfields than just a URL, mainly for not being able to load the image (alt text for HTML). We don't really need that.

Recast/iTunesPodcasts/Models/iTunes/iTunesNamespace.swift
doctor3w: uncommented, we actually want that.

Recast/SearchExampleViewController.swift
doctor3w: Oh this file is just an example so people can see how to use it. Those '!' are actually unnecessary and the items is all optional, so we will want our views to be able to handle these cases

doctor3w:
Made changes.

mindylou (dismissed):

Recast/iTunesPodcasts/Models/RSS/Enclosure.swift
```diff
@@ -0,0 +1,59 @@
...
+            let type = attributes["type"],
+            let url = URL(string: attributes["url"] ?? ""),
+            let length = Int(attributes["length"] ?? "") else {
+                return nil
+        }
+        
+        let audio_types = ["audio/x-m4a", "audio/mpeg"]
+        let video_types = ["video/quicktime", "video/mp4", "video/x-m4v"]
+        let pdf_types = ["application/pdf"]
+        if audio_types.contains(type)  {
```
jackthmp: is it better to replace this with a switch? ex:
```
switch type:
  case _ where audio_types.contains(type):
```

jackthmp (dismissed):
LGTM, just a small style change

Recast/iTunesPodcasts/Models/RSS/Enclosure.swift
```diff
@@ -13,31 +13,37 @@ public enum Enclosure {
...
+        switch type {
+        case .m4a, .mpeg:
+            self = .audio(url: url, length: length, type: type)
+        case .quicktime, .mp4, .m4v:
+            self = .video(url: url, length: length, type: type)
+        case .pdf:
             self = .pdf(url: url, length: length)
-            return
         }
         return nil
```
jackthmp: this `return nil` will always be executed, you sure this is what we want from our init? or should that be in `case default:`

jackthmp (changes requested):

Recast/iTunesPodcasts/Models/RSS/Enclosure.swift
doctor3w: ahh nah I need to insert regular returns

jackthmp (dismissed):

mindylou (dismissed):

mindylou (approved):

PR #463, opened 2018-11-28
Added MPNowPlayingInfo for player

doctor3w:
Adds lock screen now playing info and commands. There are bugs that need to be solved once we make player live forever.

### DIFF
--- a/Recast/AppDelegate.swift
+++ b/Recast/AppDelegate.swift
@@ -59,12 +59,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             let textAttributes = [NSAttributedString.Key.foregroundColor: UIColor.white]
             navController.navigationBar.titleTextAttributes = textAttributes
             navController.navigationBar.largeTitleTextAttributes = textAttributes
-            
 
             self.window = UIWindow()
             self.window?.rootViewController = navController
             self.window?.makeKeyAndVisible()
-
         }
 
         // Fabric
@@ -80,6 +78,27 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     @objc func beginInterruption() {
         // TODO: handle audio interruptions
+        // need global player
+    }
+
+    // handles headphone events
+    override func remoteControlReceived(with event: UIEvent?) {
+        super.remoteControlReceived(with: event)
+        if let e = event, e.type == .remoteControl {
+            switch e.subtype {
+            case .remoteControlPlay:
+//                Player.sharedInstance.play()
+                break
+            case .remoteControlPause, .remoteControlStop:
+//                Player.sharedInstance.pause()
+                break
+            case .remoteControlTogglePlayPause:
+//                Player.sharedInstance.togglePlaying()
+                break
+            default:
+                break
+            }
+        }
     }
 
     func applicationWillResignActive(_ application: UIApplication) {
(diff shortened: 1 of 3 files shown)


Recast/AppDelegate.swift
```diff
@@ -80,6 +78,27 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     @objc func beginInterruption() {
         // TODO: handle audio interruptions
+        // need global player
```
mindylou, line 81: what do you mean by global player?

Recast/AppDelegate.swift
```diff
@@ -80,6 +78,27 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+    }
+
+    // handles headphone events
+    override func remoteControlReceived(with event: UIEvent?) {
+        super.remoteControlReceived(with: event)
+        if let e = event, e.type == .remoteControl {
+            switch e.subtype {
+            case .remoteControlPlay:
+//                Player.sharedInstance.play()
```
mindylou, line 90: why are these commented out?

Recast/PlayerViewController.swift
```diff
@@ -328,14 +330,79 @@ class PlayerViewController: UIViewController {
...
-        // TODO: Implement
+    // Configures the Remote Command Center for our player. Should only be called once (in init)
+    func configureCommands() {
+        UIApplication.shared.beginReceivingRemoteControlEvents()
+        let commandCenter = MPRemoteCommandCenter.shared()
+        commandCenter.pauseCommand.addTarget(self, action: #selector(playPauseButtonWasPressed(_:)))
+        commandCenter.playCommand.addTarget(self, action: #selector(playPauseButtonWasPressed(_:)))
+        commandCenter.skipForwardCommand.addTarget(self, action: #selector(skipForwardButtonWasPressed(_:)))
+        commandCenter.skipBackwardCommand.addTarget(self, action: #selector(skipBackButtonWasPressed(_:)))
+        commandCenter.skipForwardCommand.preferredIntervals = [30]
```
mindylou, line 340: do we want to consider making 30 a constant?

Recast/PlayerViewController.swift
```diff
@@ -328,14 +330,79 @@ class PlayerViewController: UIViewController {
...
     }
 
     func updateNowPlayingInfo() {
-        if let imageUrl = current?.podcast?.artworkUrl600, episodeImageView != nil {
-            episodeImageView.kf.setImage(with: imageUrl)
+        guard let episode = current, let podcast = episode.podcast else {
+            configureNowPlaying(info: nil)
+            return
+        }
+        print("Updated")
```
mindylou: remove print

jackthmp (dismissed):
LGTM after you make mindy's changes

Recast/AppDelegate.swift
doctor3w: Need player to live forever, currently just lives when you open it

Recast/AppDelegate.swift
doctor3w: Need global player, we don't have singleton anymore, so reference to the view controller

Recast/PlayerViewController.swift
doctor3w: I was hoping that we'd make it some sort of setting, so for now I think just 30 is fine.

